PR #4, opened 2022-06-07
feat: Implement Server.RemoveUser

jameshoulahan:


internal/pchan/pchan.go
```diff
@@ -66,26 +68,38 @@ func (ch *PChan[T]) Push(val T, withPrio ...int) chan struct{} {
...
+
 	case <-ch.done:
+		// ...
+	}
+
+	ch.lock.Lock()
+	defer ch.lock.Unlock()
+
+	if len(ch.items) == 0 {
+		return *new(T), false
```
LBeernaertProton, line 91: Any particular reason why we can't return nil here?

jameshoulahan: Here we are returning a value of type T, which may or may not be a pointer (it's generic, we don't know what it could be). `nil` only works for pointer-like things (pointers, slices, maps, channels). So we return the default-initialized value of type T by creating a new object of type T on the heap and dereferencing it.

jameshoulahan: But yes, this is bad, shouldn't allocate on the heap just to copy back to the stack... will change to stack-allocated variable.

jameshoulahan: Done, better now

PR #6, opened 2022-06-09
feat(GODT-1619): Per connection meta data storage

LBeernaertProton:


internal/remote/operation.go
```diff
@@ -5,9 +5,19 @@ import (
 	"encoding/gob"
 )
 
+type operationMetaData struct {
```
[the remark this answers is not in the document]
jameshoulahan: Actually, i think this type isn't used at all. Can it be safely removed?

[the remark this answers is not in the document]
LBeernaertProton: This type is actually no longer required and has been removed.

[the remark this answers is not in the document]
jameshoulahan: Oops i commented before refreshing haha. Glad we came to the same conclusion.

internal/remote/conn_metadata.go
```diff
@@ -0,0 +1,102 @@
...
+
+	return buf.Bytes(), nil
+}
+
+func (c *connMetadataStore) UnmarshalBinary(data []byte) error {
+	if err := gob.NewDecoder(bytes.NewReader(data)).Decode(&c.data); err != nil {
+		return err
+	}
+
+	return nil
```
jameshoulahan, line 87:
```suggestion
	return gob.NewDecoder(bytes.NewReader(data)).Decode(&c.data)
```

sad that golangci-lint doesn't support go 1.18 properly yet as it would've complained about this. No need for a branch here.

internal/remote/conn_metadata.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+func validateKey(key any) {
```
jameshoulahan, line 90: keys in your map are only ever gonna be strings, meaning they'll always be non-nil and comparable. No need for validation.

LBeernaertProton: Fixed

internal/remote/user.go
```diff
@@ -108,35 +128,90 @@ func (user *User) send(update imap.Update, withBlock ...bool) {
...
+
+	if err := gob.NewEncoder(buf).Encode(usd); err != nil {
+		return nil, err
+	}
+
+	return buf.Bytes(), nil
+}
+
+func (usd *userSerializedData) loadFromBytes(data []byte) error {
+	if err := gob.NewDecoder(bytes.NewReader(data)).Decode(usd); err != nil {
```
jameshoulahan, line 188: Same here regarding unnecessary branch

internal/session/handle_login.go
```diff
@@ -21,7 +21,7 @@ func (s *Session) handleLogin(ctx context.Context, tag string, cmd *proto.Login,
 		return response.Bad(tag).WithError(ErrAlreadyAuthenticated)
 	}
 
-	state, err := s.backend.GetState(cmd.GetUsername(), cmd.GetPassword())
+	state, err := s.backend.GetState(ctx, cmd.GetUsername(), cmd.GetPassword(), s.sessionID)
```
jameshoulahan, line 24: unnecessary context here and in `(*user).newState(ctx context.Context, ...)`

LBeernaertProton: Fixed

internal/remote/user.go
LBeernaertProton: Fixed

jameshoulahan (approved):

PR #10, opened 2022-06-14
fix: Do not delete ConnMetadata store with SELECT or EXAMINE

LBeernaertProton:
This patch adds `State.cleanupConnMetadata()` to cleanup the metadata
rather than relying on `State.close()`. If present in the letter the
state is also cleaned up when the SELECT or EXAMINE commands are
executed, which is not what we want.

### DIFF
--- a/internal/backend/state.go
+++ b/internal/backend/state.go
@@ -466,15 +466,19 @@ func (state *State) updateMessageID(oldID, newID string) error {
 	return nil
 }
 
+func (state *State) deleteConnMetadata() error {
+	if err := state.remote.DeleteConnMetadataStore(state.metadataID); err != nil {
+		return fmt.Errorf("failed to delete conn metadata store: %w", err)
+	}
+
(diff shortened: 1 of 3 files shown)


internal/backend/state.go
```diff
@@ -466,17 +466,21 @@ func (state *State) updateMessageID(oldID, newID string) error {
 	return nil
 }
 
+func (state *State) cleanupConnMetadata() error {
```
jameshoulahan, line 469:
```suggestion
func (state *State) deleteConnMetadata() error {
```

i'd prefer this as it's more clear what "clean up" actually means.

Alternatively, just to make me happy (sorry!), consider this instead:
```suggestion
func (state *State) cleanUpConnMetadata() error {
```

LBeernaertProton: Changed to `deleteConnMetadata`

jameshoulahan (approved):

PR #12, opened 2022-06-16
feat GODT-1619:  IMAP ID extension

LBeernaertProton:


option.go
```diff
@@ -56,3 +58,17 @@ func (opt withLogger) config(server *Server) {
...
+
+type withVersionInfo struct {
+	versionInfo internal.VersionInfo
+}
+
+func (vi *withVersionInfo) config(server *Server) {
+	server.versionInfo = vi.versionInfo
+}
+
+func WithVersionInfo(info internal.VersionInfo) Option {
```
jameshoulahan, line 70: The version info cannot be internal or bridge won't be able to use it (can't import internal packages from outside the module).

You could just make it part of the top level gluon package?

imap/id.go
```diff
@@ -0,0 +1,146 @@
...
+	"encoding/gob"
+	"fmt"
+	"runtime"
+	"strings"
+
+	"github.com/ProtonMail/gluon/internal"
+)
+
+const (
+	IdKeyName             = "name"
```
jameshoulahan, line 14: Always capitalize as ID everywhere

imap/id.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+func NewID() ID {
+	return ID{
+		Name:        "Unknown",
+		Version:     "Unknown",
+		OS:          "Unknown",
+		OSVersion:   "Unknown",
+		Vendor:      "Unknown",
+		SupportUrl:  "",
```
jameshoulahan, line 52: Always capitalize as URL

internal/remote/user_queue.go
```diff
@@ -73,6 +74,19 @@ func (uoq *userOpQueue) tryGetNextOp() operation {
 	return nil
 }
 
+func (user *User) newContextWithIMAPID(id ConnMetadataID) context.Context {
```
jameshoulahan, line 77: I'd prefer if the function signature here matched that of `imap.NewContextWithIMAPID(ctx, x)`, that is, it takes in a parent context and returns a child context.

You could even pass in the parent context (which would just be `context.Background()` as an argument to `user.process(...)`.

internal/response/id.go
```diff
@@ -0,0 +1,27 @@
+package response
+
+import (
+	"github.com/ProtonMail/gluon/imap"
+)
+
+type IdResponse struct {
```
jameshoulahan, line 7: Capitalization of ID

internal/session/handle_id.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/ProtonMail/gluon/internal/response"
+)
+
+func prepareServerResponse(info *internal.VersionInfo) *response.IdResponse {
+	return &response.IdResponse{
+		ID: imap.NewIMAPIDFromVersionInfo(info),
+	}
+}
+
+func (s *Session) handleIDGet(ctx context.Context, versionInfo *internal.VersionInfo, tag string, ch chan response.Response) error {
```
jameshoulahan, line 19: Why pass in the version info as a parameter when it's already available in the session struct as a data member?

imap/id.go
```diff
@@ -0,0 +1,146 @@
...
+	writeIfNotEmpty(IdKeyEnvironment, id.Environment)
+
+	for k, v := range id.Other {
+		writeIfNotEmpty(k, v)
+	}
+
+	return fmt.Sprintf("(%v)", strings.Join(values, " "))
+}
+
+func NewIMAPIDFromKeyMap(m *map[string]string) ID {
```
jameshoulahan, line 90: map types are already pointer-like. No need to pass them in as a pointer, that's not very idiomatic and really doesn't gain you any performance.

internal/session/handle_id.go
```diff
@@ -0,0 +1,42 @@
...
+}
+
+func (s *Session) handleIDSet(ctx context.Context, versionInfo *internal.VersionInfo, tag string, cmd *proto.IDSet, ch chan response.Response) error {
+	// Update session information
+	s.imapID = imap.NewIMAPIDFromKeyMap(&cmd.Keys)
+
+	// Not logged in or no mailbox selected
+	if s.state != nil {
+		if err := s.state.SetConnMetadataKeyValue(imap.IMAPIDConnMetadataKey, s.imapID); err != nil {
+			ch <- response.Bad(tag, fmt.Sprintf("Failed to store IMAP ID: %v", err))
```
jameshoulahan, line 33: You can just directly return the bad response, no need to send it and then return an error afterwards (see eg [here](https://github.com/ProtonMail/gluon/blob/36e208671b623c7c99fe6765cb9d802cc764553c/internal/session/handle_login.go#L21))

tests/id_test.go
```diff
@@ -0,0 +1,44 @@
...
+
+	"github.com/ProtonMail/gluon/internal/response"
+
+	"github.com/ProtonMail/gluon/imap"
+	"github.com/stretchr/testify/require"
+)
+
+func TestIdNilNoData(t *testing.T) {
+	runOneToOneTest(t, "user", "pass", "/", func(c *testConnection, s *testSession) {
+		idResponse := response.IdResponse{ID: imap.NewIMAPIDFromVersionInfo(&TestServerVersionInfo)}
```
jameshoulahan, line 14: Why did you decide to make this response type not use a constructor function thing like the others?

internal/response/id.go
```diff
@@ -0,0 +1,27 @@
...
+
+func (id *IdResponse) String() string {
+	return "* ID " + id.ID.String()
+}
+
+func (r *IdResponse) Send(session Session) error {
+	return session.WriteResponse(r.String())
+}
+
+type IdNilResponse struct{}
```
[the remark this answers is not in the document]
LBeernaertProton: This type is a valid response, in case we don't wish to send any ID data back. It's not in use currently, but it is part of the extension. I can remove it if you'd like.

internal/session/handle_id.go
LBeernaertProton: :facepalm:

internal/session/handle_id.go
LBeernaertProton: Done

imap/id.go
LBeernaertProton: Fixed.

internal/session/handle_id.go
LBeernaertProton: Fixed.

internal/response/id.go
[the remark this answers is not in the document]
LBeernaertProton: Removed.

internal/remote/user_queue.go
LBeernaertProton: Fixed.

imap/id.go
LBeernaertProton: Fixed.

imap/id.go
LBeernaertProton: Fixed.

option.go
LBeernaertProton: Fixed.

internal/session/session.go
```diff
@@ -58,16 +60,23 @@ type Session struct {
...
-		caps:      []imap.Capability{imap.IMAP4rev1, imap.IDLE, imap.UNSELECT, imap.UIDPLUS, imap.MOVE},
-		sessionID: sessionID,
-		eventCh:   eventCh,
+		conn:              conn,
+		liner:             liner.New(conn),
+		backend:           backend,
+		caps:              []imap.Capability{imap.IMAP4rev1, imap.IDLE, imap.UNSELECT, imap.UIDPLUS, imap.MOVE},
+		sessionID:         sessionID,
+		eventCh:           eventCh,
+		serverVersionInfo: versionInfo,
```
jameshoulahan, line 79: Two things: long name... could just call it "version"? And also, might be nice to include this (at least the major/minor/patch) in the greeting (see `greet()` below)

LBeernaertProton: Done

jameshoulahan (approved):

PR #21, opened 2022-06-22
Fix (GODT-1637) : Forbid rename of folder when reusing the old name in the new hierarchy

rlejeune74:


internal/backend/state.go
```diff
@@ -168,7 +168,11 @@ func (state *State) Rename(ctx context.Context, oldName, newName string) error {
 			if exists, err := tx.Mailbox.Query().Where(mailbox.Name(superior)).Exist(ctx); err != nil {
 				return err
 			} else if exists {
-				continue
+				if superior == oldName {
+					return ErrExistingMailbox
+				} else {
+					continue
```
jameshoulahan, line 174: Idiomatic go style advice: if you `return` inside the `if` block, don't use an `else` block (since the code that follows will only be executed when the condition was not met).

tests/rename_test.go
```diff
@@ -2,6 +2,11 @@ package tests
 
 import (
 	"testing"
+	"time"
+
+	"github.com/emersion/go-imap"
```
jameshoulahan, line 7: More go style advice: imports should be grouped in two clusters separated by a newline, the first cluster should be stdlib stuff, the second cluster should be community stuff. Seems you've added these imports with unnecessary newlines around them.

PR #24, opened 2022-06-24
fix(GODT-1597): Handle permanently deleted messages

LBeernaertProton:
This patch ensures that messages that have been marked as deleted by the
connector are evicted from the store and the user database. To do this,
the messages table has new field named `Deleted` which tracks whether
the connector has deleted this message.

On user creation we erase all messages from the database and the store
that are marked as such. On client logout we attempt to delete all
marked messages that are not in use by other sessions which share the
same user.

Finally this patch extends the store interface with `Delete()` and the
respective implementations for the on disk and memory storage.

### DIFF
--- a/internal/backend/ent/runtime.go
+++ b/internal/backend/ent/runtime.go
@@ -4,6 +4,7 @@ package ent
 
 import (
 	"github.com/ProtonMail/gluon/internal/backend/ent/mailbox"
+	"github.com/ProtonMail/gluon/internal/backend/ent/message"
 	"github.com/ProtonMail/gluon/internal/backend/ent/schema"
 	"github.com/ProtonMail/gluon/internal/backend/ent/uid"
 )
@@ -26,6 +27,12 @@ func init() {
 	mailboxDescSubscribed := mailboxFields[4].Descriptor()
 	// mailbox.DefaultSubscribed holds the default value on creation for the Subscribed field.
 	mailbox.DefaultSubscribed = mailboxDescSubscribed.Default.(bool)
+	messageFields := schema.Message{}.Fields()
+	_ = messageFields
+	// messageDescDeleted is the schema descriptor for Deleted field.
+	messageDescDeleted := messageFields[7].Descriptor()
+	// message.DefaultDeleted holds the default value on creation for the Deleted field.
+	message.DefaultDeleted = messageDescDeleted.Default.(bool)
 	uidFields := schema.UID{}.Fields()
 	_ = uidFields
 	// uidDescDeleted is the schema descriptor for Deleted field.
--- a/internal/backend/message_tx.go
+++ b/internal/backend/message_tx.go
@@ -359,3 +359,30 @@ func txUpdateMessageID(ctx context.Context, tx *ent.Tx, oldID, newID string) err
 
 	return nil
 }
+
+func txMarkMessageAsDeleted(ctx context.Context, tx *ent.Tx, messageID string) error {
+	if _, err := tx.Message.Update().Where(message.MessageID(messageID)).SetDeleted(true).Save(ctx); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func txDeleteMessages(ctx context.Context, tx *ent.Tx, messageIDs ...string) error {
+	if _, err := tx.Message.Delete().Where(message.MessageIDIn(messageIDs...)).Exec(ctx); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func txGetMessageIDsMarkedDeleted(ctx context.Context, tx *ent.Tx) ([]string, error) {
+	messages, err := tx.Message.Query().Where(message.Deleted(true)).All(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	return xslices.Map(messages, func(t *ent.Message) string {
+		return t.MessageID
+	}), nil
+}
--- a/internal/backend/state.go
+++ b/internal/backend/state.go
@@ -7,6 +7,8 @@ import (
 	"strings"
 	"sync"
 
+	"github.com/sirupsen/logrus"
+
 	"github.com/ProtonMail/gluon/imap"
 	"github.com/ProtonMail/gluon/internal/backend/ent"
 	"github.com/ProtonMail/gluon/internal/backend/ent/mailbox"
@@ -487,6 +489,10 @@ func (state *State) deleteConnMetadata() error {
 }
 
 func (state *State) close(ctx context.Context, tx *ent.Tx) error {
+	if err := state.deleteUnusedMessagesMarkedDeleted(ctx, tx, state); err != nil {
+		logrus.WithError(err).Errorf("Failed to delete unused messages marked for delete")
+	}
+
 	state.snap = nil
 
 	state.res = nil
--- a/internal/backend/user.go
+++ b/internal/backend/user.go
@@ -26,7 +26,7 @@ type user struct {
 	stateID    int
 }
 
-func newUser(userID string, client *ent.Client, remote *remote.User, store store.Store, delimiter string) (*user, error) {
+func newUser(ctx context.Context, userID string, client *ent.Client, remote *remote.User, store store.Store, delimiter string) (*user, error) {
 	if err := client.Schema.Create(context.Background()); err != nil {
 		return nil, err
 	}
@@ -38,6 +38,10 @@ func newUser(userID string, client *ent.Client, remote *remote.User, store store
 		delimiter: delimiter,
 		client:    client,
 		states:    make(map[int]*State),
+	}
+
+	if err := user.deleteAllMessagesMarkedDeleted(ctx); err != nil {
+		return nil, err
 	}
 
 	go func() {
@@ -67,6 +71,8 @@ func (user *user) tx(ctx context.Context, fn func(tx *ent.Tx) error) error {
 	}
 
 	if err := fn(tx); err != nil {
+		logrus.WithError(err).Error("Failed to execute transaction")
+
 		if rerr := tx.Rollback(); rerr != nil {
 			panic(rerr)
 		}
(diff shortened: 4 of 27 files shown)


internal/backend/ent/runtime.go
```diff
@@ -26,6 +27,12 @@ func init() {
 	mailboxDescSubscribed := mailboxFields[4].Descriptor()
 	// mailbox.DefaultSubscribed holds the default value on creation for the Subscribed field.
 	mailbox.DefaultSubscribed = mailboxDescSubscribed.Default.(bool)
+	messageFields := schema.Message{}.Fields()
+	_ = messageFields
```
rlejeune74, line 31: what is the mistery of this empty assignment ?

internal/backend/user.go
```diff
@@ -93,3 +99,20 @@ func (user *user) close(ctx context.Context) error {
 
 	return nil
 }
+
+func (user *user) deleteAllMessagesMarkedDeleted() error {
+	ctx := context.Background()
```
jameshoulahan, line 104: Best not to "hide" internal uses of context. If an operation needs context, prefer passing it down from as high in the call chain as possible. In this case, I would suggest adding `ctx` to `backend.AddUser` and `newUser`. If someone using the backend wants to add a user, and doing so triggers some potentially expensive DB operation, they'll want to be able to stop the operation by cancelling a context.

internal/backend/message_tx.go
```diff
@@ -359,3 +359,30 @@ func txUpdateMessageID(ctx context.Context, tx *ent.Tx, oldID, newID string) err
...
+	return nil
+}
+
+func txDeleteMessagesMarkedDeleted(ctx context.Context, tx *ent.Tx, messageIDs ...string) error {
+	if _, err := tx.Message.Delete().Where(message.And(message.MessageIDIn(messageIDs...), message.Deleted(true))).Exec(ctx); err != nil {
+		return err
+	}
+
+	return nil
+}
```
jameshoulahan, line 377: These are very long lines (go style usually limits col width to 120). Further, the `Where(...)` performs `And` by default, no need to explicitly call the `And` operator.

```suggestion
func txDeleteMessagesMarkedDeleted(ctx context.Context, tx *ent.Tx, messageIDs ...string) error {
	if _, err := tx.Message.Delete().Where(
		message.MessageIDIn(messageIDs...),
		message.Deleted(true),
	).Exec(ctx); err != nil {
		return err
	}

	return nil
}
```

internal/backend/user_state.go
```diff
@@ -26,6 +26,44 @@ func (user *user) newState(metadataID remote.ConnMetadataID) (*State, error) {
...
+				if st == state || st.snap == nil {
+					continue
+				}
+				if st.snap.hasMessage(id) {
+					return false
+				}
+			}
+
+			return true
+		})
```
jameshoulahan, line 57: We already have `user.hasStateInMailboxWithMessage`, which returns whether an active state exists which is selected in the given mailbox and contains the given message. It would be useful to introduce a similar method `user.hasStateWithMessage`, which would return whether an active state exists which contains the given message (regardless of which mailbox is selected):

```go
func (user *user) hasStateWithMessage(messageID string) bool {
	user.statesLock.RLock()
	defer user.statesLock.RUnlock()

	return xslices.CountFunc(maps.Values(user.states), func(state *State) bool {
		return state.snap != nil && state.snap.hasMessage(messageID)
	}) > 0
}
```

That would allow you to simplify this block:

```suggestion
		messageIDsToDelete := xslices.Filter(messageIDsMarkForDelete, func(messageID string) bool {
			return user.hasStateWithMessage(messageID)
		})
```

store/disk.go
```diff
@@ -109,3 +110,22 @@ func (c *onDiskStore) Update(oldID, newID string) error {
...
 }
+
+func (c *onDiskStore) Delete(ids ...string) error {
+	if c.sem != nil {
+		c.sem.Lock()
+		defer c.sem.Unlock()
+	}
+
+	for _, id := range ids {
+		if err := os.Remove(filepath.Join(c.path, hashString(id))); err != nil {
```
jameshoulahan, line 121: `os.RemoveAll` is cleaner here as it returns `nil` if the path doesn't exist (avoids explicitly handling that).

internal/backend/state.go
```diff
@@ -289,6 +289,11 @@ func (state *State) Close(ctx context.Context) error {
 	return state.closeState(ctx, state)
 }
 
+func (state *State) Logout(ctx context.Context) error {
```
jameshoulahan, line 292: I don't like the fact that this is only called on logout. Clients sometimes just drop the connection (it's permitted by RFC); actually sending a logout command before disconnecting is just a nicety. 

I'd suggest we pruned the deleted messages whenever a session is closed.

internal/backend/message_tx.go
```diff
@@ -359,3 +359,30 @@ func txUpdateMessageID(ctx context.Context, tx *ent.Tx, oldID, newID string) err
...
+
+func txMarkMessageAsDeleted(ctx context.Context, tx *ent.Tx, messageID string) error {
+	if _, err := tx.Message.Update().Where(message.MessageID(messageID)).SetDeleted(true).Save(ctx); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func txDeleteMessagesMarkedDeleted(ctx context.Context, tx *ent.Tx, messageIDs ...string) error {
```
jameshoulahan, line 371: It took me a while to figure out why these functions have the signature they do, and i'm not a fan tbh. If i call something called `txDeleteMessagesMarkedDeleted`, i expect it to do exactly that, not limited just to a set of messages i then pass in.

Why complicate things? I'd just have a getter (`txGetMessagesPendingDeletion`) and setter (`txSetMessagePendingDeletion`) for the `Deleted` field, as well as a method for removing rows in the `Messages` table by ID (`txDeleteMessages`). Much simpler imo.

internal/session/handle_logout.go
```diff
@@ -14,6 +15,12 @@ func (s *Session) handleLogout(ctx context.Context, tag string, cmd *proto.Logou
 	s.capsLock.Lock()
 	defer s.capsLock.Unlock()
 
+	if s.state != nil {
```
jameshoulahan, line 18: Already mentioned this before, but just for completeness: ideally, at least the way I understand RFC, there should be no logout-specific code, as clients can theoretically drop at any time. Any logout/cleanup code should be done in `Session.done` and things that it calls (e.g. `state.Close`).

tests/deletion_pool_test.go
```diff
@@ -1,7 +1,11 @@
 package tests
 
 import (
+	"sync"
 	"testing"
+
+	"github.com/ProtonMail/gluon"
+	events2 "github.com/ProtonMail/gluon/events"
```
jameshoulahan, line 8: why call this events2? I don't see a clash anywhere.

tests/deletion_pool_test.go
```diff
@@ -358,3 +362,151 @@ func TestNoopReceivesPendingDeletionUpdates(t *testing.T) {
 		c.Sx(`A002 OK`)
 	})
 }
+
+type eventWaiter struct {
```
jameshoulahan, line 366: given we created a db_utils.go to hold just one function, i think this eventWaiter and its methods deserve their own file.

rlejeune74 (approved):

tests/session_test.go
```diff
@@ -78,6 +85,19 @@ func (s *testSession) newClient() *client.Client {
...
+		if client, err := ent.Open(dialect.SQLite, path); err == nil {
+			fn(client, context.Background())
+			return client.Close()
+		} else {
+			return err
+		}
+	}
+
+	return fmt.Errorf("User not found")
+}
```
jameshoulahan, line 99: go has shit error handling so the standard practise that's evolved is to keep the happy path out of `if` blocks and to handle errors in the `if` blocks.

The idiomatic way to write this method would then be:
```suggestion
func (s *testSession) withUserDB(user string, fn func(client *ent.Client, ctx context.Context)) error {
	path, ok := s.userDBPaths[s.userIDs[user]]
	if !ok {
		return fmt.Errorf("user not found")
	}

	client, err := ent.Open(dialect.SQLite, path)
	if err != nil {
		return err
	}

	fn(client, context.Background())

	return client.Close()
}
```

internal/backend/user_state.go
LBeernaertProton: While your example looks shorter, we have to lock/unlock countless of times. This version, while less condensed, only acquires the locks once.

LBeernaertProton: While the code changes you propose do lead to smaller code. We are forced to lock/unlock for every message id. This approach while less elegant, leads to less locking.

jameshoulahan: I'd prefer readability and maintainability here over raw performance.

LBeernaertProton: It's usually bad design and a source of contention. Right now we are not deleting many messages, but if you start deleting 100s of message, this will lead to contention and can affect other areas of the code that also interact with this lock. 

I can't comment on whether this is with Go, but most RWLock implementations from the OS favor either reading or writing, this could also lead to starvation in that case.

Since we have to access all the states, it's best to lock it once and perform our query. 

In terms of readability, I can move the filter section into a separate function, as I now realized the lock scope can be reduced.

internal/backend/message_tx.go
LBeernaertProton: This function has been replaced with txDeleteMessages().

internal/backend/state.go
LBeernaertProton: Addressed. Moved this section to `state.close()`

internal/backend/message_tx.go
LBeernaertProton: We now only have the function `txDeleteMessages`

internal/session/handle_logout.go
LBeernaertProton: As commented in another comment this whole thing has been moved to `state.close()`

tests/deletion_pool_test.go
LBeernaertProton: IDE auto-completion. Fixed.

tests/deletion_pool_test.go
LBeernaertProton: Fair point. Fixed.

tests/session_test.go
LBeernaertProton: Thanks for the suggestion. Fixed.

store/disk.go
LBeernaertProton: Thanks for pointing that out. Fixed.

PR #38, opened 2022-07-08
Fix/deadlock and demo

LBeernaertProton:
Fix demo server exit and a deadlock when IMAP clients disconnect suddenly.

### DIFF
--- a/internal/session/session.go
+++ b/internal/session/session.go
@@ -113,7 +113,12 @@ func (s *Session) Serve(ctx context.Context) error {
 		return err
 	}
 
-	return s.serve(ctx, s.getCommandCh(ctx))
+	if err := s.serve(ctx, s.getCommandCh(ctx)); err != nil {
+		logrus.WithError(err).Errorf("Failed to serve session %v", s.sessionID)
+		return err
+	}
+
+	return nil
 }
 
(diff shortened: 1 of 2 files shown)


internal/session/session.go
```diff
@@ -153,8 +158,14 @@ func (s *Session) serve(ctx context.Context, cmdCh <-chan command) error {
...
 
 		default:
-			for res := range s.handleOther(withStartTime(ctx, time.Now()), tag, cmd) {
+			responseCh := s.handleOther(withStartTime(ctx, time.Now()), tag, cmd)
+			for res := range responseCh {
 				if err := res.Send(s); err != nil {
+					// Consume all remaining channel response since the connection is no longer available.
+					// Failing to do so can cause a deadlock in the program as `s.handleOther` never finishes
+					// executing and can hold onto a number of locks indefinitely.
+					for _ = range responseCh {
```
rlejeune74, line 167: could be good to add a TODO here to track it on Jira and make sure we can't find a better way to handle the error

rlejeune74 (approved):

internal/session/session.go
LBeernaertProton: I'll create a ticket in JIRA just in case.

PR #43, opened 2022-07-13
fix(GODT-1729): Fix header parsing of fields without space

LBeernaertProton:
Fix the header parsing of fields that do not have a space after the
colon.

### DIFF
--- a/rfc822/header.go
+++ b/rfc822/header.go
@@ -270,15 +270,23 @@ func mergeMultiline(line []byte) string {
 	var res [][]byte
 
 	for line := range forLines(bufio.NewReader(bytes.NewReader(line))) {
-		res = append(res, bytes.TrimSpace(line))
+		trimmed := bytes.TrimSpace(line)
+		if trimmed != nil {
+			res = append(res, trimmed)
+		}
 	}
 
 	return string(bytes.Join(res, []byte(" ")))
 }
 
-// TODO: What if there isn't a space after the colon?
 func splitLine(line []byte) [][]byte {
-	return bytes.SplitN(line, []byte(`: `), 2)
+	result := bytes.SplitN(line, []byte(`:`), 2)
+
+	if len(result) > 1 && len(result[1]) > 0 && result[1][0] == ' ' {
+		result[1] = result[1][1:]
+	}
+
+	return result
 }
 
 // TODO: Don't assume line ending is \r\n. Bad.
--- a/tests/fetch_body_test.go
+++ b/tests/fetch_body_test.go
@@ -706,6 +706,76 @@ func TestFetchBodyPeekInvalidSectionNumber(t *testing.T) {
 	})
 }
 
+func TestFetchBody_Dovecot_InvalidMessageHeader(t *testing.T) {
+	// This tests fails when requesting when fetch BODY.PEEK[HEADER.FIELDS (In-Reply-To In-Reply-To Cc)].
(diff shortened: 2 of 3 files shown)


tests/fetch_body_test.go
```diff
@@ -657,7 +657,7 @@ Errors-to: dovecot-bounce@procontrol.fi
 X-original-sender: tss@iki.fi
 Precedence: bulk
 X-list: dovecot
-X-UID: 3                                                  
+X-UID: 3
```
cuthix, line 660: why remove whitespace in the end of the line? I think it's part of the test isn't it?

tests/fetch_body_test.go
```diff
@@ -706,6 +706,76 @@ func TestFetchBodyPeekInvalidSectionNumber(t *testing.T) {
...
+Return-Path: <tss@iki.fi>
+X-Original-To: dovecot@procontrol.fi
+Delivered-To: dovecot@procontrol.fi
+Received: from oma.irssi.org (ip213-185-36-189.laajakaista.mtv3.fi [213.185.36.189])
+	by danu.procontrol.fi (Postfix) with ESMTP id 2A7282387F
+	for <dovecot@procontrol.fi>; Mon,  7 Apr 2003 01:27:38 +0300 (EEST)
+Received: by oma.irssi.org (Postfix, from userid 1000)
+	id 3D1565E01F95; Mon,  7 Apr 2003 01:27:36 +0300 (EEST)
+Subject: [dovecot] Re: message order reversed on copying
+From: Timo Sirainen <tss@iki.fi>
```
cuthix, line 723: I am not 100% sure but this should be also valid header. 

```
From: "Timo =?UTF-8?B?VGhpcyBpcyBmdW5ueSBzdG9yeSBidXQgaXQgaXMgcG9zc2libGUgdG8gcHV0IHF1aXRlIGEgbG9u
ZyB0ZXh0IGludG8gdGhlIHN1YmplY3QgZmllbGQuLiBzbyBsb25nIHRoYXQgYWN0dWFsbHkgbmVl
ZGVkIHRvIGxpbmUgYnJlYWsgaW5zaWRlIHRoZSBxdW90YXRpb25zLg==?= Sirainen" <tss@iki.fi>
```

The idea is that when quotation is opened you don't need to put `\t` in new lines. I know it is used for long attachment names in MIME header. If it is applicable for `From:` field I am not sure.

rfc822/header.go
```diff
@@ -270,15 +270,23 @@ func mergeMultiline(line []byte) string {
 	var res [][]byte
 
 	for line := range forLines(bufio.NewReader(bytes.NewReader(line))) {
-		res = append(res, bytes.TrimSpace(line))
+		trimmed := bytes.TrimSpace(bytes.Trim(line, "\t"))
```
cuthix, line 273: isn't TrimSpace also trimming `\t` ?

tests/fetch_body_test.go
LBeernaertProton: Whitspace trimming madness. Will revert.

rfc822/header.go
LBeernaertProton: Looks like you are correct. The change must have masked some  other issue I was having at the time. Good catch.

tests/fetch_body_test.go
LBeernaertProton: Fixed.

cuthix (approved):

PR #45, opened 2022-07-14
Fix/GODT-1729 fix search issues

LBeernaertProton:
Collection of fixes related to searching queries with dates.

### DIFF
--- a/internal/backend/mailbox_search.go
+++ b/internal/backend/mailbox_search.go
@@ -392,6 +392,8 @@ func (m *Mailbox) matchSearchKeySentBefore(ctx context.Context, candidates []*sn
 			return false, err
 		}
 
+		date = convertToDateWithoutTZ(date)
+
 		return date.Before(beforeDate), nil
 	})
 }
@@ -418,8 +420,10 @@ func (m *Mailbox) matchSearchKeySentOn(ctx context.Context, candidates []*snapMs
 			return false, err
 		}
 
-		// GOMSRV-99: Is this correct?
-		return onDate.Truncate(24 * time.Hour).Equal(date.Truncate(24 * time.Hour)), nil
+		// GODT-1598: Is this correct?
+		date = convertToDateWithoutTZ(date)
+
+		return date.Equal(onDate), nil
 	})
 }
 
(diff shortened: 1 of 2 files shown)


tests/search_test.go
```diff
@@ -311,6 +332,23 @@ func TestSearchSentBefore(t *testing.T) {
...
+	// mail.ParseDate the date was being converted to 17 Feb 2003 22:29:37 +000, causing the search to pass
+	// rather than fail.
+	runOneToOneTestWithAuth(t, defaultServerOptions(t), func(c *testConnection, _ *testSession) {
+		c.doAppend("INBOX", "Date: 18 Feb 2003 00:29:37 +0200\n\nTo: foo@foo.com\r\n")
+		c.C(`A002 SELECT INBOX`)
+		c.Se(`A002 OK [READ-WRITE] SELECT`)
+
+		t := time.Now()
+
+		c.C(fmt.Sprintf(`A001 search since %s sentbefore 18-Feb-2003`, t.Format("_2-Jan-2006")))
```
cuthix, line 346: so what happens if message has 

`Date: 18 Feb 2003 22:29:37 -0200`

and I ask

`search sentbefore 18-Feb-2003`

it should return empty search again, right? 

But for both dates `18 Feb 2003 00:29:37 +0200` and `Date: 18 Feb 2003 22:29:37 -0200` 
the search

 `search sentbefore 19-Feb-2003`
 
 is non empty, right? 
 
 Is it worth to add those cases?

internal/backend/mailbox_search.go
```diff
@@ -181,7 +181,7 @@ func (m *Mailbox) matchSearchKeyBcc(ctx context.Context, candidates []*snapMsg,
 			return false, err
 		}
 
-		return strings.Contains(value, key.GetText()), nil
+		return strings.Contains(strings.ToLower(value), strings.ToLower(key.GetText())), nil
```
cuthix, line 184: would it be worth create ?

 `func containsCaseIgnored(a,b string) bool {return strings.ToLower(a), strings.ToLower(b))`

internal/backend/mailbox_search.go
```diff
@@ -213,7 +213,7 @@ func (m *Mailbox) matchSearchKeyBody(ctx context.Context, candidates []*snapMsg,
 			return false, err
 		}
 
-		return bytes.Contains(section.Body(), []byte(key.GetText())), nil
+		return bytes.Contains([]byte(strings.ToLower(string(section.Body()))), []byte(strings.ToLower(key.GetText()))), nil
```
cuthix, line 216: there is `bytes.ToLower` don't need to retype twice

PR #49, opened 2022-07-19
feat(GODT-1719): Gluon benchmark framework

LBeernaertProton:
This patch adds gluon_bench in order to enable use to write benchmarks
that run gluon in the same process as the benchmark or externally. The
support for external processes will be added in a future patch.

The benchmark tool supports producing a report to stdout or a json file.
New reporters can be added as needed.

At the moment, there is only one benchmark which measures the time to
append 1000 message to a mailbox. More benchmark will follow in the
future.

Additionally, the tool also supports running multiple benchmarks for a
configurable number of consecutive times with or without persistent
state.

Finally, this patch also extends Gluon with support to profile the
execution of individual commands inside Gluon. This information is
collected by the benchmark tool to produce its reports.

### DIFF
--- a/benchmarks/gluon_bench/reporter/json_reporter.go
+++ b/benchmarks/gluon_bench/reporter/json_reporter.go
@@ -1,0 +1,24 @@
+package reporter
+
+import (
+	"encoding/json"
+	"os"
+)
+
+// JSONReporter produces a JSON data file with all the benchmark information.
+type JSONReporter struct {
+	outputPath string
+}
+
+func (j *JSONReporter) ProduceReport(reports []*BenchmarkReport) error {
+	result, err := json.Marshal(reports)
+	if err != nil {
+		return nil
+	}
+
+	return os.WriteFile(j.outputPath, []byte(result), 0o600)
+}
+
+func NewJSONReporter(output string) *JSONReporter {
+	return &JSONReporter{outputPath: output}
+}
--- a/benchmarks/gluon_bench/reporter/reporter.go
+++ b/benchmarks/gluon_bench/reporter/reporter.go
@@ -1,0 +1,31 @@
+package reporter
+
+import (
+	"time"
+
+	"github.com/ProtonMail/gluon/benchmarks/gluon_bench/utils"
+	"github.com/ProtonMail/gluon/profiling"
+)
+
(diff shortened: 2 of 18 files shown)


benchmarks/gluon_bench/reporter/json_reporter.go
```diff
@@ -0,0 +1,24 @@
...
+
+// JSONReporter produces a JSON data file with all the benchmark information.
+type JSONReporter struct {
+	outputPath string
+}
+
+func (j *JSONReporter) ProduceReport(reports []*BenchmarkReport) error {
+	result, err := json.Marshal(reports)
+	if err != nil {
+		return nil
```
xmichelo, line 16: `return err`, I guess.

LBeernaertProton: Good catch. fixed.

benchmarks/gluon_bench/reporter/reporter.go
```diff
@@ -0,0 +1,109 @@
...
+		} else {
+			statistics.Median = sortedDurations[((statistics.SampleCount + 1) / 2)]
+		}
+		statistics.Percentile90 = sortedDurations[int(math.Floor(float64(statistics.SampleCount)*(90.0/100.0)))]
+		statistics.Percentile10 = sortedDurations[int(math.Floor(float64(statistics.SampleCount)*(10.0/100.0)))]
+		var sumSquaredWithDiv float64
+		for i := 0; i < statistics.SampleCount; i++ {
+			// Dividing now rather than later or else we will trigger overflow.
+			f64Duration := float64(sortedDurations[i])
+			sumSquaredWithDiv += (f64Duration * f64Duration) / float64(statistics.SampleCount)
```
cuthix, line 70: sorry I described wrongly

```

sumsq := float64(0)
for  x := range durations {
  sumsq += x*x
}

rms = sqrt( sumsq / len(durations) )

```

LBeernaertProton: I did that initially, but the values get to high and it overflows the counter. We just do more divisions now, but the end result is the same.

cuthix (approved):

PR #54, opened 2022-07-22
UID variant benchmarks and some minor cleanup.

LBeernaertProton:


benchmarks/gluon_bench/benchmarks/parallel_seqset.go
```diff
@@ -45,23 +45,43 @@ func NewParallelSeqSetRandom(count uint32, numWorkers uint, generateIntervals, r
...
-				index := rand.Uint32() % (count - r)
-				tmp := available[index]
-				available[index] = available[count-r-1]
-				seqSet := &imap.SeqSet{}
-				seqSet.AddNum(tmp)
-				list = append(list, seqSet)
+			if generateIntervals {
+				const maxIntervalRange = uint32(40)
+				for len(available) > 0 {
+					intervalRange := rand.Uint32() % maxIntervalRange
```
xmichelo, line 57: `rand.Int()` or `rand.Int31n()` instead of modulo.

jameshoulahan (approved):

benchmarks/gluon_bench/benchmarks/parallel_seqset.go
jameshoulahan: `maxIntervalRange` is `40` here, so `rand.Uint32() % maxIntervalRange` returns a `uint32` in range `[0, 40)`.

We could equivalently do `uint32(rand.Intn(40))` here but imo the way @LBeernaertProton did it is fine.

xmichelo: `rand() %` is a bad habit, but if you're OK, I'm fine.

PR #56, opened 2022-08-03
Expunge Benchmark

LBeernaertProton:


benchmarks/gluon_bench/benchmarks/expunge.go
```diff
@@ -0,0 +1,163 @@
...
+			*flags.UIDMode)
+
+		if err != nil {
+			return err
+		}
+
+		e.seqSets = seqSets
+		for i := uint(0); i < *flags.ParallelClients; i++ {
+			e.mailboxes = append(e.mailboxes, *flags.Mailbox)
+		}
```
jameshoulahan, line 78:
```suggestion
		messageCount := status.Messages
		if messageCount == 0 {
			return fmt.Errorf("mailbox '%v' has no messages", *flags.Mailbox)
		}

		expungeCount := uint32(*expungeCountFlag)
		if expungeCount == 0 {
			expungeCount = messageCount / 2
		}

		e.seqSets = NewParallelSeqSetExpunge(
			expungeCount,
			*flags.ParallelClients,
			*flags.RandomSeqSetIntervals,
			*flags.UIDMode,
		)

		for i := uint(0); i < *flags.ParallelClients; i++ {
			e.mailboxes = append(e.mailboxes, *flags.Mailbox)
		}
```

Few things in this code so just making a big suggestion:
- seems there's a useless `if err != nil { ... }` check (prob because `NewParallelSeqSetExpunge` used to return an error?)
- directly assign `e.seqSets` rather than creating a temp variable first (prob left over with the extra error check)
- if you break a function call over multiple lines, it's more idiomatic go style to give each param its own line, followed by the closing `)`

LBeernaertProton: Good catch. Addressed.

jameshoulahan (approved):

PR #57, opened 2022-08-03
Search benchmark

LBeernaertProton:
Adds search benchmarks for TEXT and SINCE queries. Data can be generated
randomly or fed via a new line separated text file.

### DIFF
--- a/benchmarks/gluon_bench/main.go
+++ b/benchmarks/gluon_bench/main.go
@@ -22,6 +22,8 @@ var benches = []benchmarks.Benchmark{
 	benchmarks.NewMove(),
 	benchmarks.NewStore(),
 	benchmarks.NewExpunge(),
+	benchmarks.NewSearchText(),
+	benchmarks.NewSearchSince(),
 }
 
 func main() {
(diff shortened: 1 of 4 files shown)


benchmarks/gluon_bench/main.go
```diff
@@ -22,6 +22,8 @@ var benches = []benchmarks.Benchmark{
 	benchmarks.NewMove(),
 	benchmarks.NewStore(),
 	benchmarks.NewExpunge(),
+	benchmarks.NewSearchText(),
```
jameshoulahan, line 25: I think this will lead to a _lot_ of code duplication, since every combination of search keywords will need an additional benchmark type, with its own very similar setup/teardown... e.g. it's important to be able to benchmark complex search queries with multiple terms, and it'd be annoying having to create a new fixture for each combination of terms.

This isn't a perfect solution, but perhaps you could create a single search fixture type which accepts keywords to search over on construction? e.g. `NewSearch(SearchText(), SearchSince())`. And you could maybe even create aggregators, like `NewSearch(SearchOr(SearchText(), SearchSince()))`? Might be better.

LBeernaertProton: I have some refactoring planned up ahead to streamline implementation details such as this. If it's alright with you I would like to address this in the future.

jameshoulahan: Sure, happy to wait :)

jameshoulahan (approved):

PR #65, opened 2022-08-08
fix(GODT-1618): properly handle empty LIST pattern

jameshoulahan:
RFC3501 states:
    An empty ("" string) mailbox name argument is a special request to
    return the hierarchy delimiter and the root name of the name given
    in the reference. The value returned as the root MAY be the empty
    string if the reference is non-rooted or is an empty string.

We didn't return the empty string for non-rooted references.

### DIFF
--- a/internal/backend/match.go
+++ b/internal/backend/match.go
@@ -67,28 +67,8 @@ func getMatches(
 
 // GOMSRV-100: validate this implementation.
 func match(ref, pattern, del, mailboxName string) (string, bool) {
-	// An empty ("" string) mailbox name argument is a special request to
-	// return the hierarchy delimiter and the root name of the name given
-	// in the reference. The value returned as the root MAY be the empty
-	// string if the reference is non-rooted or is an empty string.
 	if pattern == "" {
-		var res string
-
-		if strings.HasPrefix(ref, del) {
-			res += del
-		}
-
-		split := strings.Split(ref, del)
-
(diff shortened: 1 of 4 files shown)


LBeernaertProton (commented):
Looks good, but could you also add an example test for the case:

- Create mailboxes `foo` and `foo.bar` using `.` as a separator
- Run query `LIST "foo.bar" ""`, expect result `* LIST "." foo.`

The above was tested on dovecot.

jameshoulahan:
> Looks good, but could you also add an example test for the case:
> 
> * Create mailboxes `foo` and `foo.bar` using `.` as a separator
> * Run query `LIST "foo.bar" ""`, expect result `* LIST "." foo.`
> 
> The above was tested on dovecot.

If i'm not mistaken, that's already covered by `TestListRef`, specifically at `list_test.go:100`:
```
func TestListRef(t *testing.T) {
	runOneToOneTestWithAuth(t, defaultServerOptions(t, withDelimiter(".")), func(c *testConnection, s *testSession) {
		c.C(`tag create some`).OK(`tag`)
		c.C(`tag create some.thing`).OK(`tag`)
		
		...

		c.C(`tag list "some.thing" ""`).S(`* LIST (\Noselect) "." "some."`).OK(`tag`)
	})
}
```

LBeernaertProton (approved):

PR #66, opened 2022-08-09
fix(GODT-1618): Only match INBOX case-insensitively

jameshoulahan:
The only mailbox which should be treated case-insensitively is INBOX.
All other mailbox names must be matched in their original case.
This applies to all commands and to all levels of mailbox hierarchy.


internal/backend/state.go
```diff
@@ -507,3 +581,13 @@ func (state *State) close(ctx context.Context, tx *ent.Tx) error {
 
 	return nil
 }
+
+func canon(name, del string) string {
```
LBeernaertProton, line 585: This can be simplified to just check the first 5 characters/runes of the string and replace them with the uppercase version if they match the sequence `INBOX`. Alternatively, one could also do this in the parser and then you don't have check all the current and future cases.

PR #68, opened 2022-08-10
feat(GODT-1570): Badger store implementation

LBeernaertProton:
Add implementation of Badger (https://github.com/dgraph-io/badger) as
an alternative cache implementation. This will become useful later as
part of GODT-1642 where we need to ensure that cache modifications can
be rolled back with failed db modifications and vice-versa.

The unit tests have been updated to use this new implementation and it
is also available to gluon-bench using the `-store=badger` option.

Finally the `StoreBuilder` has been updated to accept a slice of bytes
rather than a string as the encryption key.


store/badger.go
```diff
@@ -0,0 +1,177 @@
...
+		return options.WithLoggingLevel(badger.ERROR)
+	default:
+		return options.WithLoggingLevel(badger.ERROR)
+	}
+}
+
+func NewBadgerStore(path string, userID string, encryptionPassphrase []byte) (*BadgerStore, error) {
+	directory := filepath.Join(path, userID)
+	db, err := badger.Open(logrusLevelToBadgerLevel(badger.DefaultOptions(directory)).
+		WithLogger(&badgerLogger{}).
```
jameshoulahan, line 60:
```suggestion
		WithLogger(logrus.StandardLogger()).
```

then you can get rid of the custom badger logger type above

jameshoulahan (changes requested):

store/badger.go
```diff
@@ -0,0 +1,177 @@
...
+		WithEncryptionKey(encryptionPassphrase).
+		WithIndexCacheSize(128 * 1024 * 1024))
+
+	if err != nil {
+		return nil, nil
+	}
+
+	store := &BadgerStore{db: db}
```
jameshoulahan, line 68: can we create `gcExitCh` here?

and in `func (b *BadgerStore) Close() error`, close the channel. Then the GC goroutine will exit (reads on closed channels always succeed), and you don't need to create/close the channel in the `startGCCollector` function

store/badger.go
LBeernaertProton: Fixed.

store/badger.go
LBeernaertProton: Good call. Fixed.

jameshoulahan (approved):

PR #72, opened 2022-08-11
fix(GODT-1618): Deal with INBOX in parser

jameshoulahan:


internal/parser/src/visitor.cpp
```diff
@@ -177,11 +184,19 @@ antlrcpp::Any Visitor::visitLogin(imap::IMAPParser::LoginContext* ctx) {
...
 }
 
 antlrcpp::Any Visitor::visitMboxOther(imap::IMAPParser::MboxOtherContext* ctx) {
-  return visit(ctx->astring());
+  auto mbox = visit(ctx->astring()).as<std::string>();
+
+  if (mbox.size() < 5)
+    return mbox;
+
+  if (strcasecmp(mbox.substr(0, mbox.find(this->mDel)).c_str(), "INBOX") == 0)
```
LBeernaertProton, line 196: Prefer to use  the `strcasencmp` and `_strnicmp` variants to avoid creating a substring to check against.

internal/parser/src/visitor.cpp
```diff
@@ -177,11 +184,19 @@ antlrcpp::Any Visitor::visitLogin(imap::IMAPParser::LoginContext* ctx) {
...
+    return "INBOX" + mbox.substr(5);
```
LBeernaertProton, line 197: Rather then discarding the old string, we could just modify the characters in place:

```
for (it = mbox.begin(); it < mbox.begin() + 5; it++) {
   *it = toupper(*it)
}

return mbox.
```

internal/backend/match.go
```diff
@@ -71,7 +71,7 @@ func match(ref, pattern, del, mailboxName string) (string, bool) {
 		return matchRoot(ref, del)
 	}
 
-	rx := fmt.Sprintf("^%v", regexp.QuoteMeta(ref+pattern))
+	rx := fmt.Sprintf("^%v", regexp.QuoteMeta(canon(ref+pattern, del)))
```
LBeernaertProton, line 74: Do you still need this here?

LBeernaertProton (commented):
It would be great if you could remove the binary artifacts from your MR since they will be re-generated by the build bot.

internal/backend/match.go
jameshoulahan: Yes I do, the IMAP grammar unfortunately doesn't treat `pattern` here as a mailbox, meaning the parser won't handle the case for us. This also allows us to handle the edge case `tag list "INB" "ox"` properly.

internal/parser/src/visitor.cpp
jameshoulahan: good idea, done

internal/parser/src/visitor.cpp
jameshoulahan: improved

LBeernaertProton (approved):

LBeernaertProton (approved):

PR #78, opened 2022-08-17
Remove TempID + Strong Typing + Remove OpQueue

LBeernaertProton:


internal/backend/state_actions.go
```diff
@@ -2,39 +2,72 @@ package backend
...
+	return DBGetMailboxByID(ctx, tx.Client(), internalID)
+}
+
+func (state *State) actionCreateMailbox(ctx context.Context, tx *ent.Tx, name string) error {
+	internalID, res, err := state.remote.CreateMailbox(ctx, state.metadataID, strings.Split(name, state.delimiter))
+	if err != nil {
+		return err
+	}
+
+	return DBCreateMailboxIfNotExists(ctx, tx, internalID, res, state.delimiter)
```
jameshoulahan, line 56: Can you use this function in `actionCreateAndGetMailbox` as well?

internal/backend/responders.go
```diff
@@ -13,18 +13,15 @@ type responder interface {
...
-	getMessageID() string
-
-	// setMessageID sets the message ID that the responder targets.
-	setMessageID(string)
+	getMessageID() imap.InternalMessageID
 }
 
 type exists struct {
-	messageID  string
+	messageID  MessageIDPair
```
jameshoulahan, line 20: I'm a bit worried that this is a `MessageIDPair` but that the `setMessageID` is gone -- it's possible that we might add a message to the snapshot with the wrong remote ID, due to a race condition, if this responder is popped from the queue, then the remote ID changes, and then we handle the responder.

Could you perhaps just have the internal ID here and query the remote ID from the DB while handling the responder?

internal/backend/snapshot_messages.go
```diff
@@ -9,7 +9,7 @@ import (
 
 // snapMsg is a single message inside a snapshot.
 type snapMsg struct {
-	ID    string
+	ID    MessageIDPair
```
jameshoulahan, line 12: I have a feeling we don't actually need the ID pair here, i can't think of a reason why the snapshot should care about the remote ID... can you?

if we could remove the remote ID here, that would simplify things even more -- can ignore my previous comment about the race condition, and we could remove the need to update remote IDs in the snapshot.

internal/backend/state_actions.go
```diff
@@ -46,91 +79,132 @@ func (state *State) actionUpdateMailbox(ctx context.Context, tx *ent.Tx, mboxID,
...
+func (state *State) actionCreateMessage(ctx context.Context, tx *ent.Tx, mboxID MailboxIDPair, literal []byte, flags imap.FlagSet, date time.Time) (int, error) {
+	internalID, res, err := state.remote.CreateMessage(ctx, state.metadataID, mboxID.RemoteID, literal, flags, date)
 	if err != nil {
 		return 0, err
 	}
 
 	update := imap.NewMessagesCreated()
 
-	if err := update.Add(res, literal, []string{mboxID}); err != nil {
+	if err := update.Add(res, literal, []imap.LabelID{mboxID.RemoteID}); err != nil {
```
jameshoulahan, line 90:
```suggestion
	if err := update.Add(res, literal, mboxID.RemoteID); err != nil {
```

i feel like `update.Add` should take the label IDs as a variadic parameter, would be nicer than casting to `[]imap.LabelID`

internal/backend/user.go
```diff
@@ -107,6 +109,8 @@ func (user *user) deleteAllMessagesMarkedDeleted(ctx context.Context) error {
 			return err
 		}
 
-		return user.store.Delete(ids...)
+		return user.store.Delete(xslices.Map(ids, func(i imap.InternalMessageID) string {
```
jameshoulahan, line 112: i think `store` should have a strongly typed interface too, since it operates on internal message IDs, do you agree?

internal/backend/state_actions.go
```diff
@@ -46,91 +79,132 @@ func (state *State) actionUpdateMailbox(ctx context.Context, tx *ent.Tx, mboxID,
...
 		return nil, err
 	}
 
-	return state.applyMessagesAddedToMailbox(ctx, tx, mboxID, messageIDs)
+	return state.applyMessagesAddedToMailbox(ctx, tx, mboxID.InternalID, internalIDs)
 }
 
-func (state *State) actionRemoveMessagesFromMailbox(ctx context.Context, tx *ent.Tx, messageIDs []string, mboxID string) error {
-	haveMessageIDs, err := DBGetMailboxMessageIDs(ctx, tx.Client(), mboxID)
+func (state *State) actionRemoveMessagesFromMailbox(ctx context.Context, tx *ent.Tx, messageIDs []MessageIDPair, mboxID MailboxIDPair) error {
```
jameshoulahan, line 171: does this really need to be ID pair as well? we could probably remove messages just by internalID?

also what about other actions, like `actionAddMessageFlags` -- why do we need both internal and remote ID there?

internal/backend/snapshot_messages.go
LBeernaertProton: It's mostly there to deal with caching data from the database. This can be replaced with a db lookup once we get closer to the connector for the actual request.

internal/backend/state_actions.go
LBeernaertProton: Good point. Will Fix.

internal/backend/user.go
LBeernaertProton: :+1:

internal/backend/state_actions.go
LBeernaertProton: In this case we need access to the created `ent.Mailbox` data. That function does not return any mailbox on purpose to avoid loading the data from disk if not required.

There is some code duplication, but they achieve different things.

internal/backend/user.go
LBeernaertProton: Done

internal/backend/state_actions.go
LBeernaertProton: Done

internal/backend/responders.go
LBeernaertProton: Done

imap/update_message_updated.go
```diff
@@ -29,8 +27,8 @@ func NewMessageUpdated(messageID string, mailboxIDs []string, seen, flagged bool
 func (u *MessageUpdated) String() string {
 	return fmt.Sprintf(
 		"MessageUpdated: MessageID = %v, MailboxIDs = %v, seen = %v, flagged = %v",
-		utils.ShortID(u.MessageID),
-		xslices.Map(u.MailboxIDs, func(id string) string { return utils.ShortID(id) }),
+		u.MessageID.ShortID(),
+		xslices.Map(u.MailboxIDs, func(id LabelID) string { return id.ShortID() }),
```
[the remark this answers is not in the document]
LBeernaertProton: Done

internal/backend/state_actions.go
LBeernaertProton: Actually, the Remove/Add Messages needs The remote ID to process the request on the connector

LBeernaertProton: The same is also applicable for the flag variants as they end up invoking `Connector.SetMessagesSeen` and `Connector.SetMessagesFlagged`

internal/backend/state_actions.go
jameshoulahan: fair enough, i wonder if there's a way to improve this (e.g. by having this function return a bool indicating if the mailbox was actually created), but let's forget this for now

jameshoulahan (approved):

PR #80, opened 2022-08-23
Poll remote connector before handling Select

jameshoulahan:


connector/connector.go
```diff
@@ -57,11 +57,7 @@ type Connector interface {
 	// GetUpdates returns a stream of updates that the gluon server should apply.
 	GetUpdates() <-chan imap.Update
 
-	// Pause pauses the stream of updates.
-	Pause()
-
-	// Resume resumes the stream of updates.
-	Resume()
+	Poll()
```
LBeernaertProton, line 60: Can you add a comment on what `Poll()` is expected to do and how it interacts with the imap.Update channel from `GetUpdates()`

LBeernaertProton (changes requested):

LBeernaertProton:
As discussed offline, lets wait until we can test against a real server implementation before making these changes to make sure this is an actual issue.

PR #81, opened 2022-08-23
Database RW Split

LBeernaertProton:
Ensure all the database queries are accessed through the scope of RWLock
    to avoid the infamous "database locked" error and to allow more parallel
    read only requests from clients.
    
 The RWLock guarantees that only one goroutine/thread can write to the
 datbase, therefore ensuring we never trigger the limitations of SQLite
 DB which causes the error.
    
 The removal of the global TX Lock now enables clients to perform more
 read only request in parallel.
    
 Going forward it is expected all database access to be accessed via the
 `DB` type.

### DIFF
--- a/internal/backend/db.go
+++ b/internal/backend/db.go
@@ -24,19 +24,41 @@ func (d *DB) Init(ctx context.Context) error {
 }
 
 func (d *DB) Read(ctx context.Context, fn func(context.Context, *ent.Client) error) error {
-	d.lock.RLock()
-	defer d.lock.Unlock()
+	_, err := DBReadResult(ctx, d, func(ctx context.Context, client *ent.Client) (struct{}, error) {
+		return struct{}{}, fn(ctx, client)
+	})
 
-	return fn(ctx, d.db)
+	return err
 }
 
 func (d *DB) Write(ctx context.Context, fn func(context.Context, *ent.Tx) error) error {
-	d.lock.Lock()
-	defer d.lock.Unlock()
+	_, err := DBWriteResult(ctx, d, func(ctx context.Context, tx *ent.Tx) (struct{}, error) {
+		return struct{}{}, fn(ctx, tx)
+	})
 
-	tx, err := d.db.Tx(ctx)
+	return err
+}
+
+func (d *DB) Close() error {
+	return d.db.Close()
+}
+
+func DBReadResult[T any](ctx context.Context, db *DB, fn func(context.Context, *ent.Client) (T, error)) (T, error) {
+	db.lock.RLock()
+	defer db.lock.RUnlock()
+
+	return fn(ctx, db.db)
+}
+
+func DBWriteResult[T any](ctx context.Context, db *DB, fn func(context.Context, *ent.Tx) (T, error)) (T, error) {
+	db.lock.Lock()
+	defer db.lock.Unlock()
+
+	var failResult T
+
+	tx, err := db.db.Tx(ctx)
 	if err != nil {
-		return err
+		return failResult, err
 	}
 
 	defer func() {
@@ -49,23 +71,20 @@ func (d *DB) Write(ctx context.Context, fn func(context.Context, *ent.Tx) error)
 		}
 	}()
 
-	if err := fn(ctx, tx); err != nil {
+	result, err := fn(ctx, tx)
+	if err != nil {
 		if rerr := tx.Rollback(); rerr != nil {
-			return fmt.Errorf("rolling back transaction: %w", rerr)
+			return failResult, fmt.Errorf("rolling back transaction: %w", rerr)
 		}
 
-		return err
+		return failResult, err
 	}
 
 	if err := tx.Commit(); err != nil {
-		return fmt.Errorf("committing transaction: %w", err)
+		return failResult, fmt.Errorf("committing transaction: %w", err)
 	}
 
-	return nil
-}
-
-func (d *DB) Close() error {
-	return d.db.Close()
+	return result, nil
 }
 
 func (b *Backend) newDB(userID string) (*DB, error) {
--- a/internal/backend/db_mailbox.go
+++ b/internal/backend/db_mailbox.go
@@ -166,7 +166,7 @@ func DBGetMailboxMessageIDPairs(ctx context.Context, client *ent.Client, mailbox
 	}), nil
 }
 
-func DBGetUIDInterval(ctx context.Context, mbox *ent.Mailbox, begin, end int) ([]*ent.UID, error) {
+func DBGetUIDInterval(ctx context.Context, client *ent.Client, mbox *ent.Mailbox, begin, end int) ([]*ent.UID, error) {
 	return mbox.QueryUIDs().
 		Where(uid.UIDGTE(begin), uid.UIDLTE(end)).
 		WithMessage().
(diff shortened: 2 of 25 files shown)


internal/backend/db_mailbox.go
```diff
@@ -166,7 +166,7 @@ func DBGetMailboxMessageIDPairs(ctx context.Context, client *ent.Client, mailbox
 	}), nil
 }
 
-func DBGetUIDInterval(ctx context.Context, mbox *ent.Mailbox, begin, end int) ([]*ent.UID, error) {
+func DBGetUIDInterval(ctx context.Context, client *ent.Client, mbox *ent.Mailbox, begin, end int) ([]*ent.UID, error) {
```
jameshoulahan, line 169: need client here?

internal/backend/db_mailbox.go
```diff
@@ -211,15 +211,15 @@ func DBGetMailboxByID(ctx context.Context, client *ent.Client, id imap.InternalM
 	return client.Mailbox.Query().Where(mailbox.MailboxID(id)).Only(ctx)
 }
 
-func DBGetMailboxMessages(ctx context.Context, mbox *ent.Mailbox) ([]*ent.UID, error) {
+func DBGetMailboxMessages(ctx context.Context, client *ent.Client, mbox *ent.Mailbox) ([]*ent.UID, error) {
```
jameshoulahan, line 214: need client here too?

internal/backend/db_mailbox.go
LBeernaertProton: While we don't need to use the client object, it is there to enforce the DB type use and to make sure we have acquire the lock. Otherwise, these queries will run at any moment, bypassing the lock and causing the infamous "database locked" error.

internal/backend/db_mailbox.go
```diff
@@ -211,15 +211,15 @@ func DBGetMailboxByID(ctx context.Context, client *ent.Client, id imap.InternalM
...
 	return mbox.QueryUIDs().WithMessage().All(ctx)
 }
 
 func DBGetMailboxRecentCount(ctx context.Context, mbox *ent.Mailbox) (int, error) {
 	return mbox.QueryUIDs().Where(uid.Recent(true)).Count(ctx)
 }
 
-func DBGetMailboxMessagesForNewSnapshot(ctx context.Context, mbox *ent.Mailbox) ([]*ent.UID, error) {
+func DBGetMailboxMessagesForNewSnapshot(ctx context.Context, client *ent.Client, mbox *ent.Mailbox) ([]*ent.UID, error) {
```
jameshoulahan, line 222: or here?

LBeernaertProton: While we don't need to use the client object, it is there to enforce the DB type use and to make sure we have acquire the lock. Otherwise, these queries will run at any moment, bypassing the lock and causing the infamous "database locked" error.

internal/backend/db.go
```diff
@@ -24,19 +24,41 @@ func (d *DB) Init(ctx context.Context) error {
 }
 
 func (d *DB) Read(ctx context.Context, fn func(context.Context, *ent.Client) error) error {
-	d.lock.RLock()
-	defer d.lock.Unlock()
+	_, err := DBReadResult[struct{}](ctx, d, func(ctx context.Context, client *ent.Client) (struct{}, error) {
```
jameshoulahan, line 27:
```suggestion
	_, err := DBReadResult(ctx, d, func(ctx context.Context, client *ent.Client) (struct{}, error) {
```

internal/backend/db.go
```diff
@@ -24,19 +24,41 @@ func (d *DB) Init(ctx context.Context) error {
...
+	})
 
-	return fn(ctx, d.db)
+	return err
 }
 
 func (d *DB) Write(ctx context.Context, fn func(context.Context, *ent.Tx) error) error {
-	d.lock.Lock()
-	defer d.lock.Unlock()
+	_, err := DBWriteResult[struct{}](ctx, d, func(ctx context.Context, tx *ent.Tx) (struct{}, error) {
```
jameshoulahan, line 35:
```suggestion
	_, err := DBWriteResult(ctx, d, func(ctx context.Context, tx *ent.Tx) (struct{}, error) {
```

LBeernaertProton: Done

internal/backend/db.go
LBeernaertProton: Done

internal/backend/user.go
```diff
@@ -53,12 +53,12 @@ func newUser(ctx context.Context, userID string, db *DB, remote *remote.User, st
...
+				update := func() error {
 					defer update.Done()
-					return user.apply(context.Background(), tx, update)
-				}); err != nil {
-					logrus.WithError(err).Errorf("Failed to apply update: %v", update)
+					return user.apply(context.Background(), update)
+				}
+				if err := update(); err != nil {
+					logrus.WithError(err).Errorf("Failed to apply update: %v", err)
 				}
```
jameshoulahan, line 62: now that `user.apply` is only used for external updates, we should probably just move this deferred `update.Done()` call into `user.apply` itself. That would simplify this weirdness here.

+ the docstring on `user.apply` is out of date, there's no `source` anymore.

internal/backend/user_updates.go
```diff
@@ -15,89 +15,100 @@ import (
...
 // applyMailboxDeleted applies a MailboxDeleted update.
-func (user *user) applyMailboxDeleted(ctx context.Context, tx *ent.Tx, update *imap.MailboxDeleted) error {
-	if exists, err := DBMailboxExistsWithRemoteID(ctx, tx.Client(), update.MailboxID); err != nil {
+func (user *user) applyMailboxDeleted(ctx context.Context, update *imap.MailboxDeleted) error {
+	if exists, err := DBReadResult(ctx, user.db, func(ctx context.Context, client *ent.Client) (bool, error) {
+		return DBMailboxExistsWithRemoteID(ctx, client, update.MailboxID)
+	}); err != nil {
 		return err
-	} else if !exists {
+	} else if exists {
```
jameshoulahan, line 90: still in need of a decent morning coffee, but why did you change this from `!exists` to `exists`...? was this a bug you fixed, or did you make a mistake here?

internal/backend/user_updates.go
```diff
@@ -15,89 +15,100 @@ import (
...
-func (user *user) applyMailboxUpdated(ctx context.Context, tx *ent.Tx, update *imap.MailboxUpdated) error {
-	client := tx.Client()
-	if exists, err := DBMailboxExistsWithRemoteID(ctx, client, update.MailboxID); err != nil {
+func (user *user) applyMailboxUpdated(ctx context.Context, update *imap.MailboxUpdated) error {
+	if exists, err := DBReadResult(ctx, user.db, func(ctx context.Context, client *ent.Client) (bool, error) {
+		return DBMailboxExistsWithRemoteID(ctx, client, update.MailboxID)
+	}); err != nil {
 		return err
-	} else if !exists {
+	} else if exists {
```
jameshoulahan, line 105: same here re `!exists` -> `exists`

internal/session/handle.go
```diff
@@ -12,7 +12,7 @@ import (
 )
 
 func (s *Session) handleOther(ctx context.Context, tag string, cmd *proto.Command, profiler profiling.CmdProfiler) chan response.Response {
-	ch := make(chan response.Response)
+	ch := make(chan response.Response, 100)
```
jameshoulahan, line 15: why buffer this channel? and how'd you arrive at 100?

LBeernaertProton: If you don't buffer the channel, you can only send one item and then you need to wait for that one item to be read. If it takes some time to send the reply from gluon to the client, the command code can't proceed until gluon fetches the next response. Buffering now allows the command to execute and release all the locks sooner as it can queue up N response items without waiting for them to be sent.

The number 100 is arbitrary, I was considering large fetch request for this case.

internal/backend/user_updates.go
LBeernaertProton: Good catch. Fixed.

internal/backend/user_updates.go
LBeernaertProton: Good catch. Fixed.

internal/backend/user.go
LBeernaertProton: Fixed.

internal/session/handle.go
jameshoulahan: Good point, nice change.

In that case, i'd like the 100 to be defined as a package constant, potentially even a variable such that it can be configured (in future, not in this MR) by a gluon init option.

jameshoulahan (approved):
looks good, just make sure to define the magic number `100` as a const or var in `internal/session/init.go` in a followup PR

PR #82, opened 2022-08-24
chore: Set library channel channel buffer constant

LBeernaertProton:


internal/session/handle.go
```diff
@@ -5,14 +5,15 @@ import (
...
+	"github.com/ProtonMail/gluon/constants"
 	"github.com/ProtonMail/gluon/internal/backend"
 	"github.com/ProtonMail/gluon/internal/parser/proto"
 	"github.com/ProtonMail/gluon/internal/response"
 	"github.com/ProtonMail/gluon/profiling"
 )
 
 func (s *Session) handleOther(ctx context.Context, tag string, cmd *proto.Command, profiler profiling.CmdProfiler) chan response.Response {
-	ch := make(chan response.Response, 100)
+	ch := make(chan response.Response, constants.ChannelBufferCount)
```
jameshoulahan, line 16: Kinda wish this was configurable somehow. Could you make it a variable like the others in `internal/session/init.go`? The default value could of course be taken from `constants.ChannelBufferCount`.

LBeernaertProton: Done

jameshoulahan (approved):

PR #85, opened 2022-08-25
fix(GODT-1647): Address double call to `Connector.Close()`

LBeernaertProton:
Rework goroutine shutdown so that we don't need to call
`Connector.Close()` when closing the backend user.

Two separate channels have been added ensure the blocking go-routines
exit when the Server shuts down or an user has been removed.

### DIFF
--- a/internal/backend/user.go
+++ b/internal/backend/user.go
@@ -25,7 +25,8 @@ type user struct {
 	statesLock  sync.RWMutex
 	nextStateID int
 
-	updateWG sync.WaitGroup
+	updateWG     sync.WaitGroup
+	updateQuitCh chan struct{}
 }
 
 func newUser(ctx context.Context, userID string, db *DB, remote *remote.User, store store.Store, delimiter string) (*user, error) {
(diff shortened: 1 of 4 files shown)


internal/backend/user.go
```diff
@@ -52,9 +54,14 @@ func newUser(ctx context.Context, userID string, db *DB, remote *remote.User, st
 		defer user.updateWG.Done()
 		labels := pprof.Labels("go", "Connector Updates", "UserID", user.userID)
 		pprof.Do(ctx, labels, func(_ context.Context) {
-			for update := range remote.GetUpdates() {
-				if err := user.apply(context.Background(), update); err != nil {
-					logrus.WithError(err).Errorf("Failed to apply update: %v", err)
+			for {
+				select {
+				case update := <-remote.GetUpdates():
```
jameshoulahan, line 59: Can we call `GetUpdates` outside the `for` loop, hold the channel in a variable, and then use it inside the for loop? I don't like that we repeatedly call `GetUpdates` each iteration; the implementation of the connector might construct a new channel each call, for instance.

LBeernaertProton: That would be a terrible connector, but I see your point. Fixed.

internal/remote/user.go
```diff
@@ -57,16 +59,14 @@ func (user *User) GetUpdates() <-chan imap.Update {
 }
 
 func (user *User) Close(ctx context.Context) error {
+	close(user.forwardQuitCh)
+	user.forwardWG.Wait()
+	user.updatesCh = nil
```
jameshoulahan, line 64: this scares me a bit, because here we set `user.updatesCh` to nil, and below (in `forward(...)`) we close it. Closing a nil channel leads to a panic.

i'm trying to follow the system of waitgroups and defers to see if we're guaranteed not to close after setting it to nil but im getting a bit confused haha -- i'm thinking we don't even need to set it to nil at all? kind of a pointless operation, no?

LBeernaertProton: It is safe a this point. But removed as it is unnecessary.

jameshoulahan (approved):
Looks good. I hope we remove the `remote` package soon; all this unnecessary forwarding means we just have to take care of more goroutines...

PR #88, opened 2022-08-26
Enable race checking tests and fix all reported issues

LBeernaertProton:


internal/backend/mailbox.go
```diff
@@ -160,8 +174,13 @@ func (m *Mailbox) Copy(ctx context.Context, seq *proto.SequenceSet, name string)
 		return nil, ErrNoSuchMailbox
 	}
 
-	messages, err := m.snap.getMessagesInRange(ctx, seq)
-	if err != nil {
+	var messages []*snapMsg
+
+	if err := snapshotRead(m.snap, func(s *snapshot) error {
```
jameshoulahan, line 179: could use `snapshotReadErr` here

internal/backend/mailbox.go
```diff
@@ -198,12 +217,21 @@ func (m *Mailbox) Move(ctx context.Context, seq *proto.SequenceSet, name string)
...
 		return nil, ErrNoSuchMailbox
 	}
 
-	messages, err := m.snap.getMessagesInRange(ctx, seq)
-	if err != nil {
+	var messages []*snapMsg
+
+	var snapMBoxID MailboxIDPair
+
+	if err := snapshotRead(m.snap, func(s *snapshot) error {
```
jameshoulahan, line 229: could also use `snapshotReadErr` here

internal/backend/mailbox.go
```diff
@@ -234,8 +262,14 @@ func (m *Mailbox) Move(ctx context.Context, seq *proto.SequenceSet, name string)
...
+	var messages []*snapMsg
+
+	if err := snapshotRead(m.snap, func(s *snapshot) error {
+		var err error
+		messages, err = s.getMessagesInRange(ctx, seq)
+
+		return err
+	}); err != nil {
 		return err
 	}
```
jameshoulahan, line 274:
```suggestion
	messages, err := snapshotReadErr(m.snap, func(s *snapshot) ([]*snapMsg, error) {
		return s.getMessagesInRange(ctx, seq)
	})
	if err != nil {
		return err
	}
```

internal/backend/mailbox.go
```diff
@@ -269,13 +303,17 @@ func (m *Mailbox) Expunge(ctx context.Context, seq *proto.SequenceSet) error {
 	var msg []*snapMsg
```
jameshoulahan, line 303:
```
	msg, err := snapshotReadErr(m.snap, func(s *snapshot) ([]*snapMsg, error) {
		if seq == nil {
			return s.getAllMessages(), nil
		}

		return s.getMessagesInRange(ctx, seq)
	})
	if err != nil {
		return err
	}

	return m.expunge(ctx, msg)
```

could also make this nicer with one call to `snapshotReadErr`

internal/backend/state_actions.go
```diff
@@ -147,14 +147,16 @@ func (state *State) actionAddMessagesToMailbox(
 ) (map[imap.InternalMessageID]int, error) {
 	var haveMessageIDs []MessageIDPair
```
jameshoulahan, line 148: can remove this line; you create the variable again on line 150

internal/backend/state_actions.go
```diff
@@ -213,14 +215,16 @@ func (state *State) actionMoveMessages(
 	{
 		var messageIDsToAdd []MessageIDPair
```
jameshoulahan, line 216: can remove this line

jameshoulahan (changes requested):
I mostly have cosmetic comments (suggest using more `snapshotReadErr`). But i wonder two things...

1. sure, this mitigates the race condition where we replace the snapshot (via `SELECT`/`EXAMINE`) while we are using it (async with imap updates). But these changes just allow us to pull data _out_ of the snapshot safely. There's no guarantee the data will still be up to date by the time we use it! e.g. we might extract the snapshot's mailbox ID via a threadsafe `snapshotRead` call, then another thread replaces the snapshot, and then we continue executing using an out of date mailbox ID. What can we do about that?
2. could clean up the code a lot by providing helper methods for commonly used things fetched via the wrapper, e.g. `snapshotWrapper.getInternalID()` could provide safe access to the snapshot's internal ID without having to use the full callback approach each time

LBeernaertProton:
Regarding 1). I agree this solution won't fix that problem. But to address this properly will require more deeper refactors. I would like to look at that afterwards and split the current `State` and `user` objects up to make some things more obvious.

internal/backend/mailbox.go
LBeernaertProton: Done

internal/backend/mailbox.go
LBeernaertProton: Done

internal/backend/mailbox.go
LBeernaertProton: Done

internal/backend/mailbox.go
LBeernaertProton: Done

internal/backend/state_actions.go
LBeernaertProton: Done

internal/backend/state_actions.go
LBeernaertProton: Done

jameshoulahan (approved):

PR #89, opened 2022-08-30
Remove state & snapshot locks

LBeernaertProton:
Prior to this patch the updates to other session states were applied
immediately by going through all the available states in `backend.user`
and applying the changes immediately.

This patch changes the architecture so that all state updates for other
session states are queued via a channel rather than executed
immediately. This removes the need for snapshot lock, snapshot messages
lock, IDLE lock and responder lock as we no longer have two sessions
accessing state from one another.

To ensure the current state is still update, we mark all context with
the stateID (when available) and if the stateID matches one of the
states in the list, the update is applied immediately.

The downside is that every state now has to verify that this update can
be executed locally rather than checking in advance.

Finally this patch also fixes some update code to remove unnecessary
memory duplication.

Note: The `userWrapper` is a temporary type that will be remove in the
next patch.

### DIFF
--- a/internal/backend/state.go
+++ b/internal/backend/state.go
@@ -29,7 +29,6 @@ type State struct {
 	ro   bool
 
 	doneCh chan struct{}
-	stopCh chan struct{}
 
 	updatesQueue *queue.QueuedChannel[stateUpdate]
 
@@ -375,8 +374,6 @@ func (state *State) Done() <-chan struct{} {
 }
 
 func (state *State) Close(ctx context.Context) error {
-	defer close(state.stopCh)
-
 	return state.user.removeState(ctx, state.stateID)
 }
 
--- a/internal/backend/state.go
+++ b/internal/backend/state.go
@@ -4,42 +4,67 @@ import (
 	"context"
 	"fmt"
 	"strings"
-	"sync"
 
 	"github.com/ProtonMail/gluon/imap"
 	"github.com/ProtonMail/gluon/internal/backend/ent"
+	"github.com/ProtonMail/gluon/internal/queue"
 	"github.com/ProtonMail/gluon/internal/remote"
 	"github.com/ProtonMail/gluon/internal/response"
 	"github.com/bradenaw/juniper/sets"
 	"github.com/bradenaw/juniper/xslices"
+	"github.com/sirupsen/logrus"
 )
 
-// TODO(REFACTOR): Decide on the best way to pass around user/state/snap objects! Currently quite gross...
 type State struct {
-	*user
+	user stateUserAccessor
 
 	stateID    int
 	metadataID remote.ConnMetadataID
 
-	idleCh   chan response.Response
-	idleLock sync.RWMutex
+	idleCh chan response.Response
 
-	res     []responder
-	resLock sync.Mutex
+	res []responder
 
-	snap *snapshotWrapper
+	snap *snapshot
 	ro   bool
 
 	doneCh chan struct{}
 	stopCh chan struct{}
+
+	updatesQueue *queue.QueuedChannel[stateUpdate]
+
+	delimiter string
+}
+
+type stateUpdate interface {
+	// filter returns true when the state can be passed into apply.
+	filter(s *State) bool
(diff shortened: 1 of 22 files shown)


internal/backend/state.go
```diff
@@ -363,7 +378,31 @@ func (state *State) Done() <-chan struct{} {
...
+
+func (state *State) GetStateUpdatesCh() <-chan stateUpdate {
+	if state == nil {
+		return nil
+	}
+
+	return state.stateUpdatesQueue.GetChannel()
+}
+
+func (state *State) QueueUpdates(updates ...stateUpdate) bool {
```
jameshoulahan, line 392: can be private?

internal/session/session.go
```diff
@@ -145,6 +145,13 @@ func (s *Session) serve(ctx context.Context, cmdCh <-chan command) error {
 		case <-s.state.Done():
 			return nil
 
+		case stateUpdate := <-s.state.GetStateUpdatesCh():
```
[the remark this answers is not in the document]
jameshoulahan:
```
To ensure the current state is still update, we mark all context with
the stateID (when available) and if the stateID matches one of the
states in the list, the update is applied immediately.
```

how does that relate to my question here?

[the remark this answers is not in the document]
LBeernaertProton: We had exactly the same problem before. You can execute a valid command on the current session's state, but then a responder comes in and before you run the command changes the state of your mailbox and now the command is invalid.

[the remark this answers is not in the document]
jameshoulahan: whoops my bad
```
	// If we detect a state id in the context, it means this function call is a result of a User interaction.
	// When that happens the update needs to be applied to the state matching the state ID immediately. If no such
	// stateID exists or the context information is not present, all updates are queued for later execution.
```

just came across that in the code, ignore me ^^

internal/backend/state_updates.go
```diff
@@ -9,6 +9,48 @@ import (
...
 )
 
+type messageFlagsAddedStateUpdate struct {
+	anyMessageIDStateFilter
+	flags   imap.FlagSet
+	mboxID  MailboxIDPair
+	stateID int
+}
+
+func NewMessageFlagsAddedStateUpdate(flags imap.FlagSet, mboxID MailboxIDPair, messageIDs []imap.InternalMessageID, stateID int) stateUpdate {
```
jameshoulahan, line 19: private?

internal/backend/state_updates.go
```diff
@@ -47,30 +85,48 @@ func (state *State) applyMessageFlagsAdded(ctx context.Context, tx *ent.Tx, mess
...
-				newFlags = newFlags.Set(imap.FlagDeleted, snapFlags.Contains(imap.FlagDeleted))
-			}
+type messageFlagsRemovedStateUpdate struct {
+	anyMessageIDStateFilter
+	flags   imap.FlagSet
+	mboxID  MailboxIDPair
+	stateID int
+}
+
+func NewMessageFlagsRemovedStateUpdate(flags imap.FlagSet, mboxID MailboxIDPair, messageIDs []imap.InternalMessageID, stateID int) stateUpdate {
```
jameshoulahan, line 102: private? as with others below

LBeernaertProton: Done

internal/backend/state_updates.go
LBeernaertProton: Done

internal/backend/state.go
LBeernaertProton: Done

internal/backend/state.go
```diff
@@ -4,42 +4,68 @@ import (
...
+	res []responder
 
-	snap *snapshotWrapper
+	snap *snapshot
 	ro   bool
 
 	doneCh chan struct{}
 	stopCh chan struct{}
+
+	stateUpdatesQueue *queue.QueuedChannel[stateUpdate]
```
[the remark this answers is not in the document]
LBeernaertProton: done

jameshoulahan (approved):

PR #91, opened 2022-08-31
fix(GODT-1757): Refactor code to have better data separation

LBeernaertProton:
Due to Go's lackluster import system we had to refactor certain types
into separate packages so we can better control what each type has
access to.

All the database related code has been moved into the db package.

All the remote user management code has been been moved into the remote
package. Existing remote code has been erased.

All things related to state and snapshot have been moved into the state
package.

To avoid cyclic imports, the state package declares two interfaces which
provide access to some user specific data as well as a custom connector
implementation on which metadata can be attached.


internal/backend/user.go
```diff
@@ -102,106 +117,125 @@ func (user *user) close(ctx context.Context) error {
...
-				if !update.filter(state) {
-					return nil
-				}
-
-				return update.apply(ctx, tx, state)
-			}
-		})
+func (user *user) queueStateUpdate(update state.Update) {
+	if err := user.forState(func(state *state.State) error {
+		state.QueueUpdates(update)
```
houlahaj, line 135: Could we check the bool response here? if it's false, the queue is closed, and it'd be nice to return an error or panic, for future debugging purposes

jameshoulahan, line 135: Could we check the bool return value here? If it's false, the queue is closed, and it would be nice to return an error or panic (for future debugging purposes)

internal/backend/user.go
```diff
@@ -102,106 +117,125 @@ func (user *user) close(ctx context.Context) error {
...
 }
 
-func (s *stateUserAccessor) applyMessagesMovedFromMailbox(
-	ctx context.Context,
-	tx *ent.Tx,
-	mboxFromID, mboxToID imap.InternalMailboxID,
-	messageIDs []imap.InternalMessageID,
-) (map[imap.InternalMessageID]int, error) {
-	return s.u.applyMessagesMovedFromMailbox(ctx, tx, mboxFromID, mboxToID, messageIDs)
+func (user *user) getStates() []*state.State {
```
jameshoulahan, line 227: this is unused, should delete

internal/backend/connector_updates.go
```diff
@@ -392,41 +378,20 @@ func (user *user) applyMessagesMovedFromMailbox(
...
-			return nil, err
-		}
-	}
-
-	if _, err := DBAddMessagesToMailbox(ctx, tx, messageIDs, mboxToID); err != nil {
-		return nil, err
-	}
-
-	messageUIDs, err := DBGetMessageUIDs(ctx, tx.Client(), mboxToID, messageIDs)
+	messageUIDs, updates, err := state.MoveMessagesFromMailbox(ctx, tx, mboxFromID, mboxToID, messageIDs)
```
jameshoulahan, line 381: `applyMessagesMovedFromMailbox` is unused -- might as well delete it

internal/state/context.go
```diff
@@ -0,0 +1,26 @@
+package state
+
+import "context"
+
+type stateContextType struct{}
+
+var stateContextKey stateContextType
+
+func NewStateContext(ctx context.Context, s *State) context.Context {
```
jameshoulahan, line 9: Would appreciate a docstring describing what the state context represents and how we use it internally

tests/fetch_body_test.go
```diff
@@ -669,7 +669,7 @@ v0.95 2002-07-31  Timo Sirainen <tss@iki.fi>
 	+ Code cleanups
 	- Found several bugs from mempool and ioloop code, now we should
 	  be stable? :)
-	- A few corrections for long header field handling
+	- Apply few corrections for long header field handling
```
jameshoulahan, line 672: Expected change? Not sure why you needed to change this.

tests/parallel_test.go
```diff
@@ -13,7 +13,7 @@ import (
 func TestSelectWhileSyncing(t *testing.T) {
 	runOneToOneTestClientWithAuth(t, defaultServerOptions(t), func(client *client.Client, s *testSession) {
 		// Define some mailbox names.
-		mailboxNames := []string{"A", "B", "C", "D", "E", "F", "G", "H", "I", "J"}
+		mailboxNames := []string{"Apply", "B", "C", "D", "E", "F", "G", "H", "I", "J"}
```
jameshoulahan, line 16: Ah so maybe the `A` -> `Apply` in the test expectation came from here, more find-replace shenanigans? or was this intentional for some reason?

tests/run_test.go
```diff
@@ -28,7 +28,7 @@ func runOneToOneTestWithAuth(tb testing.TB, options *serverOptions, tests func(*
 	})
 }
 
-// runOneToOneTestWithData runs a test with one account and one connection. A mailbox is created with test data.
+// runOneToOneTestWithData runs a test with one account and one connection. Apply mailbox is created with test data.
```
jameshoulahan, line 31: Wow yes i see more `A` -> `Apply`, definitely not expected :D

tests/benchmark_fetch_test.go
```diff
@@ -17,8 +17,8 @@ func BenchmarkFetchDatabase(b *testing.B) {
 					s.messageCreatedFromFile("user", benchID, `testdata/multipart-mixed.eml`)
 				}
 
-				c.C(`A SELECT BENCH`)
-				c.Se(`A OK [READ-WRITE] (^_^)`)
+				c.C(`Apply SELECT BENCH`)
```
jameshoulahan, line 20: Apply

jameshoulahan (changes requested):
All good, just would remove a couple bits of dead code, and you were a bit heavy-handed with the find-replace :)

internal/backend/user.go
```diff
@@ -102,106 +117,125 @@ func (user *user) close(ctx context.Context) error {
...
-
-				return update.apply(ctx, tx, state)
-			}
-		})
+func (user *user) queueStateUpdate(update state.Update) {
+	if err := user.forState(func(state *state.State) error {
+		state.QueueUpdates(update)
+		return nil
+	}); err != nil {
+		panic("Unexpected, should not happen")
```
jameshoulahan, line 138: tiny style comment: errors and panics in go should begin with a lowercase. I think the linter will only complain about errors beginning with an uppercase, but we should extend this to panics too.

internal/errors/errors.go
```diff
@@ -1,4 +1,4 @@
-package backend
+package errors
```
jameshoulahan, line 1: I'm not such a fan of this package. Firstly, because it shares its name with the stdlib package `errors`. But secondly, packages should really define their own errors (unless it's a common system-wide error, such as `io.EOF`). In our case, i don't think these errors are so ubiquitous that they deserve their own package.

There's a clear split in error usage:
```
$ rg "errors\.Err"
tests/state_test.go
76:			c.Sx(fmt.Sprintf("%d NO %v", i, errors.ErrSessionNotSelected))

internal/backend/backend.go
89:		return errors.ErrNoSuchUser
132:		return errors.ErrNoSuchUser
162:	return "", errors.ErrNoSuchUser

internal/state/mailbox.go
163:		return nil, errors.ErrNoSuchMailbox
206:		return nil, errors.ErrNoSuchMailbox

internal/state/snapshot.go
55:		return 0, errors.ErrNoSuchMessage
64:		return 0, errors.ErrNoSuchMessage
73:		return nil, errors.ErrNoSuchMessage
82:		return errors.ErrNoSuchMessage
234:		return errors.ErrNoSuchMessage
242:		return errors.ErrNoSuchMailbox
252:		return errors.ErrNoSuchMessage
279:		return 0, errors.ErrNoSuchMessage
293:		return 0, errors.ErrNoSuchMessage
302:		return 0, errors.ErrNoSuchMessage

internal/state/state.go
79:		return errors.ErrNoSuchMailbox
112:		return errors.ErrNoSuchMailbox
146:			return nil, errors.ErrExistingMailbox
183:		return errors.ErrNoSuchMailbox
200:			return Result{}, errors.ErrNoSuchMailbox
206:			return Result{}, errors.ErrExistingMailbox
215:					return Result{}, errors.ErrExistingMailbox
260:				return errors.ErrNoSuchMailbox
279:		return errors.ErrNoSuchMailbox
281:		return errors.ErrAlreadySubscribed
294:		return errors.ErrNoSuchMailbox
296:		return errors.ErrAlreadyUnsubscribed
320:		return errors.ErrNoSuchMailbox
339:		return errors.ErrSessionNotSelected
347:		return errors.ErrNoSuchMailbox
```

I suggest we move these errors to `internal/state/errors.go`:
- `ErrNoSuchMessage`
- `ErrNoSuchMailbox`
- `ErrExistingMailbox`
- `ErrAlreadySubscribed`
- `ErrAlreadyUnsubscribed`
- `ErrSessionNotSelected` (which we could rename to `ErrMailboxNotSelected`)

This can go back to `internal/backend/errors.go`:
- `ErrNoSuchUser`

These two we can delete (they're not used):
- `ErrNoSuchSnapshot`
- `ErrNotImplemented`

internal/backend/backend.go
```diff
@@ -63,7 +61,7 @@ func (b *Backend) AddUser(ctx context.Context, userID string, conn connector.Con
 		return err
 	}
 
-	db, err := b.newDB(userID)
+	db, err := db.NewDB(b.dir, userID)
```
jameshoulahan, line 64: `db.NewDB` appends `/db` to the given directory before using it. Probably better to do that here (like we do on line 59 with `/store`) and keep the db package simple, thoughts?

internal/state/connector.go
```diff
@@ -0,0 +1,65 @@
+package state
+
+import (
+	"context"
+	"time"
+
+	"github.com/ProtonMail/gluon/imap"
+)
+
+type Connector interface {
```
jameshoulahan, line 10: Would be nice to have an explanation here about what this connector is for and how it differs from the externally-supplied top-level `connector.Connector`

internal/backend/user.go
LBeernaertProton: The for state loop doesn't allow for anything other than error returns.

LBeernaertProton: I'll add an error log if an update couldn't be queued. The only time the queue will be closed is during shutdown.

LBeernaertProton: Done

internal/backend/user.go
LBeernaertProton: Done

internal/state/context.go
LBeernaertProton: Done

tests/fetch_body_test.go
LBeernaertProton: This is why I have trust issues with auto refactors :O. Fixed.

tests/parallel_test.go
LBeernaertProton: Fixed.

tests/run_test.go
LBeernaertProton: Fixed.

tests/benchmark_fetch_test.go
LBeernaertProton: Fixed.

internal/backend/user.go
LBeernaertProton: I usually see this warning for `fmt.Errof`. I don't get it for panics though.

Addressed.

internal/errors/errors.go
LBeernaertProton: Noted. Fixed.

internal/backend/backend.go
LBeernaertProton: Good call. Fixed.

internal/state/connector.go
LBeernaertProton: Done

jameshoulahan (approved):

PR #93, opened 2022-09-01
fix(GODT-1586): Implement context cancellation for commands

LBeernaertProton:
Check for context cancellation for long running commands such as list,
fetch, lsub and search.


internal/session/command.go
```diff
@@ -14,39 +12,29 @@ import (
...
 			for {
 				tag, cmd, err := s.readCommand(del)
-				if err != nil {
-					if errors.Is(err, io.EOF) {
-						return
-					} else if err := response.Bad(tag).WithError(err).Send(s); err != nil {
+
+				select {
+				case cmdCh <- command{tag: tag, cmd: cmd, err: err}:
+					if err != nil {
```
jameshoulahan, line 31: should continue reading future commands here rather than erroring out

LBeernaertProton: Fixed.

jameshoulahan (dismissed):

internal/session/command.go
```diff
@@ -14,39 +12,42 @@ import (
...
-				}
 
-				switch {
-				case cmd.GetStartTLS() != nil:
-					if err := s.handleStartTLS(tag, cmd.GetStartTLS()); err != nil {
-						if err := response.Bad(tag).WithError(err).Send(s); err != nil {
+				if err == nil && cmd.GetStartTLS() != nil {
+					// TLS needs to be handled here in order to ensure that next command read is over the
+					// tls connection.
+					if e := s.handleStartTLS(tag, cmd.GetStartTLS()); e != nil {
```
jameshoulahan, line 32: If there's an error handling the STARTTLS command, why do we then go on to send the command on the cmdCh? the command will be nil

LBeernaertProton: we forward the error to be handled by the main session.

jameshoulahan (approved):

PR #96, opened 2022-09-02
GODT-1773 reporter integration

LBeernaertProton:
Add support for remote reporting and report some information of interest.

### DIFF
--- a/builder.go
+++ b/builder.go
@@ -12,6 +12,7 @@ import (
 	"github.com/ProtonMail/gluon/internal/backend"
 	"github.com/ProtonMail/gluon/internal/session"
 	"github.com/ProtonMail/gluon/profiling"
+	"github.com/ProtonMail/gluon/reporter"
 	"github.com/ProtonMail/gluon/store"
 )
 
@@ -24,6 +25,7 @@ type serverBuilder struct {
 	versionInfo        internal.VersionInfo
 	cmdExecProfBuilder profiling.CmdProfilerBuilder
 	storeBuilder       store.Builder
+	reporter           reporter.Reporter
 }
 
 func newBuilder() (*serverBuilder, error) {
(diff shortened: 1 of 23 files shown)


jameshoulahan (changes requested):
In general I really like the changes! Keen to approve. I just find the changes very verbose... it's a lot of extra code to check the context, report the error, and log failure.

How about naming the package `report`, then introducing some helper functions, for example:
```go
package report

func MessageWithContext(ctx context.Context, message string, context Context) {
	reporter, ok := GetReporterFromContext(ctx)
	if !ok {
		return
	}

	if err := reporter.ReportMessageWithContext(message, context); err != nil {
		logrus.WithError(err).Errorf("Failed to report message")
	}
}
```

The helper hides the work of checking the context for a reporter and logging failed report calls.

Then for example, instead of doing
```go
	if err := user.close(ctx); err != nil {
		if rep, ok := reporter.GetReporterFromContext(ctx); ok {
			context := reporter.Context{
				"error": err,
			}
			if re := rep.ReportMessageWithContext("Failed close user from Backend.RemoveUser()", context); re != nil {
				logrus.WithError(re).Errorf("Failed to report error to remote")
			}
		}

		return fmt.Errorf("failed to close backend user: %w", err)
	}
```

you simply do
```go
	if err := user.close(ctx); err != nil {
		report.MessageWithContext(
			ctx, 
			"Failed to close user from Backend.RemoveUser()", 
			report.Context{ "error": err, },
		)

		return fmt.Errorf("failed to close backend user: %w", err)
	}
```

LBeernaertProton:
Good point, this is why it should be forbidden to write code on Friday afternoon ;)

LBeernaertProton:
Comment addressed

jameshoulahan (approved):

PR #98, opened 2022-09-05
 fix(GODT-1593): list of capabilities based on state. 

cuthix:


LBeernaertProton:
Looks good, but please rebase and as @jameshoulahan mentioned, use the conventions listed in https://www.conventionalcommits.org/

store/badger.go
```diff
@@ -20,10 +21,23 @@ type badgerTransaction struct {
 	tx *badger.Txn
 }
 
+func sanitizePassphrase(origin []byte) []byte {
```
jameshoulahan, line 24: Rather than sanitizing the provided key, i'd prefer we just hash it with a 256-bit SHA hash so that whatever the user provides, it's gonna end up 256-bits (32 bytes) long (even for shorter passphrases).

internal/session/handle_capability.go
```diff
@@ -3,15 +3,31 @@ package session
 import (
 	"context"
 
+	"github.com/ProtonMail/gluon/imap"
 	"github.com/ProtonMail/gluon/internal/parser/proto"
 	"github.com/ProtonMail/gluon/internal/response"
 )
 
+func (s *Session) getCaps() []imap.Capability {
+	if s.state != nil {
```
jameshoulahan, line 12: normally we protect the `s.state` with `s.userLock` (which should probably be called stateLock now). Can you try to lock that here too? That might conflict with existing locks though, you might have to be a bit more clever.

LBeernaertProton: The lock is no longer necessary at the session level. Everything that happens at the session is single threaded now (barring any unexpected go-routines)

jameshoulahan: Great, then perhaps we should remove it. But we'll have to be mindful once we allow multiple in-flight commands per session to put some kind of state lock back here.

LBeernaertProton (dismissed):

jameshoulahan (approved):

PR #100, opened 2022-09-06
Untagged reponse behavior

LBeernaertProton:
This patch fixes the behavior of untagged responses to match dovecot.
All untagged responses are now sent at the end of the command before the
closing tag is sent and no longer before and after. Some of the unit
tests had to be updated to match this new behavior.

This patch also fixes fetch so that it applies the seen flag to the
snapshot on which fetch is operating. This is now possible as snapshots
aren't shared between sessions anymore.

### DIFF
--- a/internal/state/responders.go
+++ b/internal/state/responders.go
@@ -85,8 +85,12 @@ func (u *exists) handle(ctx context.Context, tx *ent.Tx, snap *snapshot) ([]resp
 	res := []response.Response{response.Exists().WithCount(seq)}
 
 	if recent := len(snap.getMessagesWithFlag(imap.FlagRecent)); recent > 0 {
-		if err := db.ClearRecentFlag(ctx, tx, snap.mboxID.InternalID, u.messageID); err != nil {
+		if msgFlags, err := snap.getMessageFlags(u.messageID); err != nil {
 			return nil, err
+		} else if msgFlags.Contains(imap.FlagRecent) {
+			if err := db.ClearRecentFlag(ctx, tx, snap.mboxID.InternalID, u.messageID); err != nil {
+				return nil, err
+			}
 		}
 
 		res = append(res, response.Recent().WithCount(recent))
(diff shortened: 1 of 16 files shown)


internal/state/responders.go
```diff
@@ -85,8 +85,12 @@ func (u *exists) handle(ctx context.Context, tx *ent.Tx, snap *snapshot) ([]resp
...
 
 	if recent := len(snap.getMessagesWithFlag(imap.FlagRecent)); recent > 0 {
-		if err := db.ClearRecentFlag(ctx, tx, snap.mboxID.InternalID, u.messageID); err != nil {
+		if msgFlags, err := snap.getMessageFlags(u.messageID); err != nil {
 			return nil, err
+		} else if msgFlags.Contains(imap.FlagRecent) {
+			if err := db.ClearRecentFlag(ctx, tx, snap.mboxID.InternalID, u.messageID); err != nil {
+				return nil, err
+			}
 		}
```
jameshoulahan, line 94:
```suggestion
		msgFlags, err := snap.getMessageFlags(u.messageID)
		if err != nil {
			return nil, err
		}

		if msgFlags.Contains(imap.FlagRecent) {
			if err := db.ClearRecentFlag(ctx, tx, snap.mboxID.InternalID, u.messageID); err != nil {
				return nil, err
			}
		}
```

interesting, would've expected the linter to complain about this

jameshoulahan (commented):
In general looks good, i just don't like how `handleCommand` returns a response to indicate forced exit, while `handleWithMailbox` returns a response that should be treated as a tagged OK -- the difference is a bit confusing. Is there a way we could unify the return values? Perhaps we could give `Session` a method `forceClose(...)` that forces it to send a `bye` response with some data at the end of the command?

internal/state/responders.go
LBeernaertProton: > In general looks good, i just don't like how `handleCommand` returns a response to indicate forced exit, while `handleWithMailbox` returns a response that should be treated as a tagged OK -- the difference is a bit confusing. Is there a way we could unify the return values? Perhaps we could give `Session` a method `forceClose(...)` that forces it to send a `bye` response with some data at the end of the command?

This change happened in a previous MR. We can discuss an improvement for a new MR as that is not the scope of this change.

PR #101, opened 2022-09-06
fix(GODT-1757): Flag updates can only be applied through responders

LBeernaertProton:
We should only apply flag updates through responders as it possible to
run into a sequence of events where an exist responder hasn't been
processed yet. Attempting to apply the flag changes during a state
update will result in an error or no op.

The error was not reproducible at the moment since the state filter
would fail to queue the update as the message did not exist yet.

Finally the whole issue is currently masked since we get the new
message for the snap shot directly from the database. I ran into this
issue with local changes where this was no longer the case.


internal/state/responders.go
```diff
@@ -181,7 +191,22 @@ func (u *fetch) handle(ctx context.Context, tx *ent.Tx, snap *snapshot) ([]respo
...
+		newMessageFlags = curFlags.AddFlagSet(u.flags)
+	case FetchFlagOpRem:
+		newMessageFlags = curFlags.RemoveFlagSet(u.flags)
+	case FetchFlagOpSet:
+		newMessageFlags = u.flags
+	}
+
+	if u.toggleDeleted {
+		newMessageFlags = newMessageFlags.Set(imap.FlagDeleted, curFlags.Contains(imap.FlagDeleted))
+	}
```
jameshoulahan, line 207: Is this actually toggling? Shouldn't it be `!curFlags.Contains(imap.FlagDeleted)`?

jameshoulahan (approved):

internal/state/responders.go
LBeernaertProton: renamed to `cameFromDifferentMailbox` for clarity.

jameshoulahan (approved):

PR #102, opened 2022-09-07
fix(GODT-1757): Recent flag behavior and snapshot db reads

LBeernaertProton:
This patch removes the need for the snapshots to read new messages from the database when they arrive via an `exists` responder.

We have also updated the behavior of the `Recent` flag to match dovecot's implementation. If a client append a message to a selected mailbox only that client will receive the recent flags. If a client appends a message to a non-selected mailbox, the first state which receives the update will apply the recent flag.

To achieve the latter, we added a new update `ExistsStateUpdate` which enforces the logic described above in a thread safe manner.

### DIFF
--- a/internal/session/session.go
+++ b/internal/session/session.go
@@ -139,6 +139,13 @@ func (s *Session) serve(ctx context.Context) error {
 
 	for {
 		select {
+		case stateUpdate := <-s.state.GetStateUpdatesCh():
+			if err := s.state.ApplyUpdate(ctx, stateUpdate); err != nil {
+				logrus.WithError(err).Error("Failed to apply state update")
+			}
+
+			continue
+
 		case res, ok := <-cmdCh:
 			if !ok {
 				logrus.Debugf("Failed to read from command channel")
@@ -162,13 +169,6 @@ func (s *Session) serve(ctx context.Context) error {
 
 		case <-s.state.Done():
 			return nil
-
-		case stateUpdate := <-s.state.GetStateUpdatesCh():
-			if err := s.state.ApplyUpdate(ctx, stateUpdate); err != nil {
-				logrus.WithError(err).Error("Failed to apply state update")
-			}
-
-			continue
 
 		case <-ctx.Done():
 			return ctx.Err()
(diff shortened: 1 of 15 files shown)


internal/state/responders.go
```diff
@@ -99,12 +112,93 @@ func (u *exists) handle(ctx context.Context, tx *ent.Tx, snap *snapshot) ([]resp
...
+
+	return &ExistsStateUpdate{
+		MBoxIDStateFilter: MBoxIDStateFilter{MboxID: mailboxID},
+		responders:        responders,
+		targetStateID:     stateID,
+	}
+}
+
+func newExistsStateUpdateWithExists(mailboxID imap.InternalMailboxID, responders []*exists, s *State) Update {
+	var stateID int
```
jameshoulahan, line 156: risky, can this potentially match the first state ID if we pass in a nil state? because i think the first state ID that the backend generates is `0`, which would be the default value here.

LBeernaertProton: We increment the stateID before using it. But you are right, could become an issue with overflow.

jameshoulahan: ah okay so it starts at 0 but the first state gets 1. Might be worth using a pointer so we avoid using in-band values here.

LBeernaertProton: I added a bool safeguard, safer than keeping around pointers which may no longer exist.

internal/session/session.go
```diff
@@ -139,6 +139,13 @@ func (s *Session) serve(ctx context.Context) error {
 
 	for {
 		select {
+		case stateUpdate := <-s.state.GetStateUpdatesCh():
+			if err := s.state.ApplyUpdate(ctx, stateUpdate); err != nil {
+				logrus.WithError(err).Error("Failed to apply state update")
+			}
+
+			continue
+
```
jameshoulahan, line 148: Any reason to move this above? just checking that you don't expect it to change behaviour

LBeernaertProton: Yes. If we move this up we can guarantee that the update will be processed before the next command comes in. I had some flaky tests because some updates were processed in between commands. This sort of fixes the issue.

jameshoulahan: As discussed, order of select branches doesn't matter. But moving it here can't harm anything 🤷

jameshoulahan (approved):

PR #104, opened 2022-09-07
Refactor: Strong types for UID UIDValidity and IMAP Sequence

LBeernaertProton:


imap/seqset.go
```diff
@@ -26,6 +27,12 @@ func (seqval SeqVal) String() string {
 
 type SeqSet []SeqVal
```
cuthix, line 28: SeqVal has two `int`s should we use `uint32` ?

internal/response/exists.go
```diff
@@ -5,14 +5,14 @@ import (
...
-	count int
+	count uint32
 }
 
 func Exists() *exists {
 	return &exists{}
 }
 
-func (r *exists) WithCount(n int) *exists {
+func (r *exists) WithCount(n uint32) *exists {
```
jameshoulahan, line 15: Technically this is also of type `imap.Seq` -- an exists response returns the sequence number of the last message in the mailbox

internal/backend/user.go
```diff
@@ -29,7 +29,7 @@ type user struct {
 
 	db *db.DB
 
-	states      map[int]*state.State
+	states      map[state.StateID]*state.State
 	statesLock  sync.RWMutex
 	nextStateID int
```
jameshoulahan, line 34: why not make this of type `state.StateID`?

imap/seqset.go
LBeernaertProton: Good catch. fixed.

internal/response/exists.go
LBeernaertProton: Fixed.

internal/backend/user.go
LBeernaertProton: Good idea. Done.

jameshoulahan (approved):
Wondering whether the state's ID should be private and only accessible via a getter, just to help enforce the user/state/backend separation we're trying to create. But I don't mind how it is now -- approved

PR #105, opened 2022-09-07
fix: Properly close listeners and connections during teardown

jameshoulahan:


server.go
```diff
@@ -218,44 +228,31 @@ func (s *Server) GetUserDataPath(userID string) (string, error) {
...
-func (s *Server) removeListener(l net.Listener) {
-	s.listenersLock.Lock()
-	defer s.listenersLock.Unlock()
+	// Wait until all goroutines currently handling connections are done.
+	s.serveWG.Wait()
 
-	if _, ok := s.listeners[l]; ok {
-		delete(s.listeners, l)
+	// Close the server error channel.
+	close(s.serveErrCh)
```
LBeernaertProton, line 243: You may want to close this after `s.backend.close()` if a Session close fails for some reason it will result in a panic when trying to write to the err channel.

LBeernaertProton (changes requested):

server.go
jameshoulahan: Good point, done

internal/queue/queued_channel.go
```diff
@@ -30,14 +43,56 @@ func NewQueuedChannel[T any](channelBufferSize int, capacity int) *QueuedChannel
...
 func (q *QueuedChannel[T]) Close() {
-	q.queue.Close()
+	q.closed.Store(closed)
+
+	q.cond.L.Lock()
+	defer q.cond.L.Unlock()
+
+	q.cond.Broadcast()
+
+	for _, item := range q.items {
```
LBeernaertProton, line 73: Leftover debug logs?

jameshoulahan: oops yeah! was leftover from when the leak detector was complaining

jameshoulahan: Fixed. Also moved queue out of internal (i want to use it in bridge too)

internal/queue/queued_channel.go
```diff
@@ -30,14 +41,52 @@ func NewQueuedChannel[T any](channelBufferSize int, capacity int) *QueuedChannel
 	return queue
 }
 
-func (q *QueuedChannel[T]) Queue(items ...T) bool {
-	return q.queue.PushMany(items...)
+func (q *QueuedChannel[T]) Enqueue(items ...T) bool {
+	if q.closed.Load().(bool) {
```
LBeernaertProton, line 45: Can we switch this back to the old code which uses the uint32 atomic functions so we don't have to do type casts?

jameshoulahan: I took the atomic.Bool implementation from Go 1.19 stdlib, i think that's the best option for now, then we switch to official atomic.Bool later

LBeernaertProton (approved):

PR #111, opened 2022-09-08
fix(GODT-1570): Improve Append command perf

LBeernaertProton:
Do not load the entire mailbox from the database if it is not selected and only flush event if the mailbox is currently selected.

### DIFF
--- a/internal/state/snapshot.go
+++ b/internal/state/snapshot.go
@@ -44,6 +44,14 @@ func newSnapshot(ctx context.Context, state *State, client *ent.Client, mbox *en
 	return snap, nil
 }
 
+func newEmptySnapshot(state *State, mbox *ent.Mailbox) *snapshot {
+	return &snapshot{
+		mboxID:   ids.NewMailboxIDPair(mbox),
+		state:    state,
+		messages: newMsgList(),
+	}
+}
+
 func (snap *snapshot) hasMessage(messageID imap.InternalMessageID) bool {
 	return snap.messages.has(messageID)
 }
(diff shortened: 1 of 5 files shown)


internal/state/snapshot.go
```diff
@@ -44,6 +44,16 @@ func newSnapshot(ctx context.Context, state *State, client *ent.Client, mbox *en
...
+func newEmptySnapshot(ctx context.Context, state *State, mbox *ent.Mailbox) (*snapshot, error) {
+	snap := &snapshot{
+		mboxID:   ids.NewMailboxIDPair(mbox),
+		state:    state,
+		messages: newMsgList(),
+	}
+
+	return snap, nil
+}
+
```
jameshoulahan, line 56:
```suggestion
func newEmptySnapshot(ctx context.Context, state *State, mbox *ent.Mailbox) *snapshot {
	return &snapshot{
		mboxID:   ids.NewMailboxIDPair(mbox),
		state:    state,
		messages: newMsgList(),
	}
}

```

LBeernaertProton: Fixed

internal/state/state.go
```diff
@@ -355,6 +355,25 @@ func (state *State) Mailbox(ctx context.Context, name string, fn func(*Mailbox)
 	return fn(newMailbox(mbox, state, snap))
 }
 
+// EmptyMailbox does not load the mailbox snapshot data from the database and passes true into the function
+// if the currently selected mailbox matches the requested mailbox.
+func (state *State) EmptyMailbox(ctx context.Context, name string, fn func(*Mailbox, bool) error) error {
```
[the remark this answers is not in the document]
LBeernaertProton: Suggestions applied. Renamed to AppendOnlyInbox.

jameshoulahan (approved):

PR #114, opened 2022-09-09
refactor(GODT-1570): Add CreateAndAddMessageToMailbox

LBeernaertProton:
Reduces the total amount of database queries for Append by having an optimized path that re-use previously available calculations.

### DIFF
--- a/internal/state/actions.go
+++ b/internal/state/actions.go
@@ -103,36 +103,27 @@ func (state *State) actionCreateMessage(
 		return 0, err
 	}
 
-	var reqs []*db.CreateMessageReq
+	msg := update.Messages[0]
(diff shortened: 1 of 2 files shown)


internal/state/actions.go
```diff
@@ -103,7 +103,7 @@ func (state *State) actionCreateMessage(
 		return 0, err
 	}
 
-	var reqs []*db.CreateMessageReq
+	var reqs db.CreateMessageReq
```
jameshoulahan, line 106: Tiny thing, but could you call this "req" since there's just one?

LBeernaertProton: Done

jameshoulahan (approved):

PR #115, opened 2022-09-09
fix: Don't close the listener when closing the server

jameshoulahan:
When calling gluon.Serve(ctx, net.Listener), we take in a listener. We shouldn't close this when closing gluon as we don't own this listener. The surrounding caller who calls gluon.Serve(...) should be responsible for closing it, to avoid double-close errors.


LBeernaertProton (changes requested):
Looks good but can you also update the demo and benchmark projects?

jameshoulahan:
> Looks good but can you also update the demo and benchmark projects?

Oops, done.

LBeernaertProton (approved):

PR #119, opened 2022-09-12
fix(GODT-1617): correctly parse headers with space lines.

cuthix:


rfc822/header.go
```diff
@@ -218,7 +218,7 @@ func ParseHeader(header []byte) *Header {
 		split := splitLine(line)
 
 		switch {
-		case len(bytes.TrimSpace(line)) == 0:
+		case len(bytes.Trim(line, "\r\n")) == 0:
```
LBeernaertProton, line 221: Shouldn't this be wrapped with `bytes.TrimSpace(bytes.Trim(line, "\r\n")) ?

rfc822/parser.go
```diff
@@ -143,7 +143,7 @@ func Split(b []byte) ([]byte, []byte, error) {
 
 		header = append(header, b...)
 
-		if len(bytes.TrimSpace(b)) == 0 {
+		if len(bytes.Trim(b, "\r\n")) == 0 {
```
LBeernaertProton, line 146: Shouldn't this be wrapped with `bytes.TrimSpace(bytes.Trim(line, "\r\n"))` ?

rfc822/header.go
cuthix: The header line which contains just a whitespace is allowed:

```
Subject: which has line with space
␣
 is allowed
````

`TrimSpace` would trim also space and tab symbols, which is the issue I tried to fix.

The correct header / body separator should be empty line `\r\n\r\n`.

PR #121, opened 2022-09-12
fix(GODT-1599): Make sure `search before/since/on` uses internal date.

cuthix:
* fix(godt-1599): make sure search before uses internal date.
* fix: more strict `testConnection.S` and `.Se` matching.
* fix(godt-1599): make sure search since/on uses internal date.

N.B. Dependency `emersion/go-mbox` was forked, added `GetMessageDelimiter` function and replaced in go.mod

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,12 +4,12 @@ go 1.18
 
 require (
 	entgo.io/ent v0.11.2
+	github.com/ProtonMail/go-mbox v1.1.0
 	github.com/ProtonMail/gopenpgp/v2 v2.4.7
 	github.com/bradenaw/juniper v0.6.0
 	github.com/dgraph-io/badger/v3 v3.2103.2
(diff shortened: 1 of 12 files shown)


go.mod
```diff
@@ -9,7 +9,7 @@ require (
 	github.com/dgraph-io/badger/v3 v3.2103.2
 	github.com/emersion/go-imap v1.2.1-0.20220429085312-746087b7a317
 	github.com/emersion/go-imap-uidplus v0.0.0-20200503180755-e75854c361e9
-	github.com/emersion/go-mbox v1.0.2
+	github.com/emersion/go-mbox v1.0.3
```
jameshoulahan, line 12: I'd suggest you just use the full dependency `github.com/ProtonMail/go-mbox` here rather than replacing it -- just change the module name in that repo to match.

jameshoulahan (approved):

PR #123, opened 2022-09-13
feat: Option to remove files when removing a gluon user

jameshoulahan:


internal/backend/backend.go
```diff
@@ -100,6 +95,16 @@ func (b *Backend) RemoveUser(ctx context.Context, userID string) error {
 
 	delete(b.users, userID)
 
+	if removeFiles {
+		if err := b.storeBuilder.Delete(filepath.Join(b.dir, "store"), userID); err != nil {
+			return err
+		}
+
+		if err := db.DeleteDB(filepath.Join(b.dir, "db"), userID); err != nil {
```
LBeernaertProton, line 103: Have you considered using a constant for the store and db path so that they remain in sync with other locations?

tests/server_test.go
```diff
@@ -154,7 +146,7 @@ func runServer(tb testing.TB, options *serverOptions, tests func(session *testSe
 			TestServerVersionInfo.Vendor,
 			TestServerVersionInfo.SupportURL,
 		),
-		gluon.WithStoreBuilder(&testBadgerStoreBuilder{}),
+		gluon.WithStoreBuilder(&store.BadgerStoreBuilder{}),
```
LBeernaertProton, line 149: The Test badger store function also suppresses the badger log system so that when one traces the test there no noise from badger. Could you please restore that?

LBeernaertProton (changes requested):
I'm thinking that perhaps it may be better to  have two separate functions `DisableUser|RemoveUser` and `RemoveUser|DeleteUser`. Internally it could just call the existing implmentation of `RemoveUser` but it might be cleaner on the API side of things?

internal/backend/backend.go
jameshoulahan: added functions `getStoreDir()` and `getDBDir()` which are used everywhere to get the paths.

jameshoulahan:
> I'm thinking that perhaps it may be better to have two separate functions `DisableUser|RemoveUser` and `RemoveUser|DeleteUser`. Internally it could just call the existing implmentation of `RemoveUser` but it might be cleaner on the API side of things?

I actually think that becomes a bit more confusing, because that API suggests that it's possible to firstly remove a user and then delete it, which wouldn't be the case -- once the user is removed, you can't then delete it. Having it in the same function (with an option for file removal) feels better to me since.

LBeernaertProton (approved):

PR #128, opened 2022-09-15
refactor: Add Event/Update base type

jameshoulahan:


LBeernaertProton (commented):
What purpose does this serve to have the `_isUpdate()` or `_isEvent()` methods?

jameshoulahan:
> What purpose does this serve to have the `_isUpdate()` or `_isEvent()` methods?

It's just a way to prevent foreign code outside those packages from defining updates. Just some rudimentary protection.

Also a reason i made those base types was to define a base `String()` method to give all updates/events a way to render themselves (but i didn't do that in this PR)

LBeernaertProton (approved):

PR #129, opened 2022-09-15
refactor(GODT-1730): Refactor HeaderType

LBeernaertProton:
To improve lookup after the header has been parsed all the entries are stored in a map.

Every single line of the header is stored as a linked list so we can perform order dependent queries or modify the header while preserving the original order. The result of mergedLine requests are cached to avoid this computation in the future.

The SetHeaderValue function has been updated so that it does not parse the full header, but rather sets the new value as the first value of the new header. This avoids having to parse the whole header to set our internal ID-Key

Finally our internal ID key was case-corrected according to the standard.

### DIFF
--- a/rfc822/header.go
+++ b/rfc822/header.go
@@ -6,105 +6,208 @@ import (
 	"fmt"
 	"io"
 	"net/textproto"
-	"regexp"
 	"strings"
-
-	"golang.org/x/exp/slices"
 )
 
-var rxWhitespace = regexp.MustCompile(`^\s+`)
+type headerEntry struct {
+	ParsedHeaderEntry
+	mapKey string
+	merged string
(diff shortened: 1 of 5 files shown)


rfc822/header.go
```diff
@@ -334,25 +484,8 @@ func NewHeaderParser(header []byte) HeaderParser {
...
-				return nil, err
-			}
-		}
-
-		lines = append(lines, header[entry.keyStart:entry.valueEnd])
-	}
-
-	return &Header{lines: lines}, nil
+func ParseHeader(data []byte) (*Header, error) {
+	return NewHeader(data)
```
jameshoulahan, line 488: Still need two separate functions?

jameshoulahan (approved):

PR #133, opened 2022-09-19
fix: Header parser (naive)

jameshoulahan:
The new header parser did not correctly deal with two edge cases:
1. The message contains an embedded RFC822 message with no header
2. Parsing something (not necessarily a header) with a key but no value

This fix checks that what we think is a header actually is one in a naive way by trying to parse it and seeing if it returns an error. It also handles the case that a header key was not finished being read by returning an error, indicating this wasn't a valid header).

### DIFF
--- a/rfc822/parser.go
+++ b/rfc822/parser.go
@@ -155,6 +155,10 @@ func Split(b []byte) ([]byte, []byte) {
 func parse(literal []byte, identifier []int, begin, end int) *Section {
 	header, _ := Split(literal[begin:end])
 
+	if !isHeader(header) {
+		header = nil
+	}
+
 	return &Section{
 		identifier: identifier,
 		literal:    literal,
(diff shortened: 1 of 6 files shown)


rfc822/parser.go
```diff
@@ -155,6 +155,10 @@ func Split(b []byte) ([]byte, []byte) {
 func parse(literal []byte, identifier []int, begin, end int) *Section {
 	header, _ := Split(literal[begin:end])
 
+	if !isHeader(header) {
```
LBeernaertProton, line 158: Can we update this piece of code so that when we have a valid header you cache the result in the section for the next time we re-use the parsed header here?

jameshoulahan: done

LBeernaertProton (approved):

PR #136, opened 2022-09-19
feat: Naive parallel fetch

jameshoulahan:


LBeernaertProton (commented):
The changes look good but:

- Have you benchmarked this with one and multiple clients?
- Are you sure the code you are parallelizing is thread safe (e.g.: setting seen flag)
- What if a mail client does many parallel fetches by itself, how do you prevent this from launching too many threads?

PR #137, opened 2022-09-19
refactor(GODT-1730): Improve boundary scanner

LBeernaertProton:
Use byte scans instead of file IO to do boundary separation.

This patch also deletes the old scanner and encrypt.go, as they are not used anymore inside gluon.


rfc822/parser.go
```diff
@@ -57,30 +57,33 @@ func (section *Section) Literal() []byte {
...
 	if len(identifier) > 0 {
-		children := section.Children()
+		children, err := section.Children()
+		if err != nil {
+			return nil, err
+		}
 
 		if identifier[0] <= 0 || identifier[0]-1 > len(children) {
-			return nil
+			return nil, nil
```
jameshoulahan, line 78: return error here? because the requested part doesn't exist

LBeernaertProton: Good point. fixed.

jameshoulahan (approved):

PR #138, opened 2022-09-19
fix(GODT-1616): close not sending unilateral updates.

cuthix:


internal/state/responders.go
```diff
@@ -254,7 +255,7 @@ func (u *expunge) handle(snap *snapshot, _ StateID) ([]response.Response, respon
 	}
 
 	// When handling a CLOSE command, EXPUNGE responses are not sent.
-	if u.asClose {
```
cuthix, line 257: Information `u.asClose` is not needed in this implementation, should I remove it completely?

LBeernaertProton: I think it should be safe to do so :thinking:

internal/state/responders.go
```diff
@@ -299,7 +300,7 @@ func NewFetch(messageID imap.InternalMessageID, flags imap.FlagSet, asUID, asSil
 	}
 }
 
-func (u *fetch) handle(snap *snapshot, _ StateID) ([]response.Response, responderDBUpdate, error) {
+func (u *fetch) handle(ctx context.Context, snap *snapshot, _ StateID) ([]response.Response, responderDBUpdate, error) {
```
jameshoulahan, line 303: Could you name the context `_` in cases where it's unused?

jameshoulahan (dismissed):

LBeernaertProton (approved):

PR #141, opened 2022-09-20
Use uint64 for internalIDs

LBeernaertProton:


internal/db/mailbox.go
```diff
@@ -241,8 +244,8 @@ func TranslateRemoteMailboxIDs(ctx context.Context, client *ent.Client, mboxIDs
 	}), nil
 }
 
-func CreateMailboxIfNotExists(ctx context.Context, tx *ent.Tx, internalID imap.InternalMailboxID, mbox imap.Mailbox, delimiter string) error {
-	exists, err := MailboxExistsWithID(ctx, tx.Client(), internalID)
+func CreateMailboxIfNotExists(ctx context.Context, tx *ent.Tx, mbox imap.Mailbox, delimiter string) error {
+	exists, err := MailboxExistsWithRemoteID(ctx, tx.Client(), mbox.ID)
```
jameshoulahan, line 248: is this changed because this used to be a bug?

internal/backend/user.go
```diff
@@ -37,22 +38,33 @@ type user struct {
 
 	// statesWG is
 	statesWG sync.WaitGroup
+
+	messageIDCounter uint64
```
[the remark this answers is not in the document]
LBeernaertProton: Since this is used with atomic instructions later on, it is better to leave this type as is to avoid issues with temporary addresses being incremented rather that the actual address.

internal/db/mailbox.go
LBeernaertProton: Since the mailbox ID's are auto generated in the database now, we don't have access to the internal ID, so the check has been replaced with the remote ID instead.

jameshoulahan (approved):

PR #143, opened 2022-09-21
Improve snapshot seq range

LBeernaertProton:


internal/state/snapshot_messages.go
```diff
@@ -119,3 +125,31 @@ func (list *snapMsgList) seq(seq imap.SeqID) (*snapMsg, bool) {
...
+	for ; index < len && list.msg[index].UID < uidLo; index++ {
+	}
+
+	start := index
+
+	// find last UID
+	for ; index < len && list.msg[index].UID <= uidHi; index++ {
+	}
+
+	return list.msg[start:index]
```
jameshoulahan, line 148: binary search would be faster given they're sorted by definition

jameshoulahan (approved):

PR #146, opened 2022-09-22
Parallel Fetchs and flag processing improvements.

LBeernaertProton:


jameshoulahan (approved):
I wonder why you picked 16 as the hard limit for min parallelism? why not just use parallel if > 1 item is fetchable? can you back this choice up with benchmarks?

LBeernaertProton:
Depending on the size of the message and the fetch query it can sometimes be faster to just do them one by one up to that point.

We can improve the decision process in the future to a case where if you need to access the literal, we lower the limit to 1. Actually, that might be a better idea. brb, writing mr.

PR #148, opened 2022-09-23
Improve Performance of COPY command

LBeernaertProton:
The following collection of changes improves COPY perf by 9-10x.

### DIFF
--- a/internal/state/actions.go
+++ b/internal/state/actions.go
@@ -152,7 +152,7 @@ func (state *State) actionAddMessagesToMailbox(
 	messageIDs []ids.MessageIDPair,
 	mboxID ids.MailboxIDPair,
 	isMailboxSelected bool,
-) (map[imap.InternalMessageID]*ent.UID, error) {
+) ([]db.UIDWithFlags, error) {
 	{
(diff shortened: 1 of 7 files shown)


internal/state/actions.go
```diff
@@ -197,21 +192,14 @@ func (state *State) actionAddMessagesToMailbox(
 	return messageUIDs, nil
 }
 
-func (state *State) actionRemoveMessagesFromMailbox(
+// actionRemoveMessagesFromMailboxUnchecked is similar to actionRemoveMessagesFromMailbox, but it does not validate
+// the input for whether messages actually exist in the database or if the message set is empty.
+func (state *State) actionRemoveMessagesFromMailboxUnchecked(
```
jameshoulahan, line 197: i'd like a comment here to describe in what cases this is actually useful -- describe why we don't need this validation in some cases

jameshoulahan (dismissed):

internal/state/actions.go
LBeernaertProton: Done.

jameshoulahan (approved):

PR #150, opened 2022-09-26
refactor(GODT-1570): Improve Examine, Status and Select + Batch flag processing

LBeernaertProton:
Avoid scanning all the messages if we only need one message or a number of messages.


jameshoulahan (approved):
Approving for now but we should revisit this later with better solution; i don't like mix-and-matching the case sensitivity of flag comparisons.

PR #151, opened 2022-09-26
refactor(GODT-1570): Do less work with Flag modification

LBeernaertProton:
Cost of retrieving deleted messages and filtering them and then updating the ones that were not set was about the same as just setting the value for every message in the list.

Return the flags as an array as there's no need to perform mapping operations.

### DIFF
--- a/internal/db/message.go
+++ b/internal/db/message.go
@@ -314,10 +314,15 @@ func GetMessageUIDsWithFlagsAfterAddOrUIDBump(ctx context.Context, client *ent.C
 	return result, nil
 }
 
+type MessageFlagSet struct {
+	ID      imap.InternalMessageID
+	FlagSet imap.FlagSet
+}
+
 // GetMessageFlags returns the flags of the given messages.
 // It does not include per-mailbox flags (\Deleted, \Recent)!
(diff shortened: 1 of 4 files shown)


internal/db/message.go
```diff
@@ -331,18 +336,21 @@ func GetMessageFlags(ctx context.Context, client *ent.Client, messageIDs []imap.
...
+			for _, v := range msg.Edges.Flags {
+				mfs.FlagSet.AddToSelf(v.Value)
+			}
 
-	for _, message := range messages {
-		curFlags[message.ID] = imap.NewFlagSetFromSlice(xslices.Map(message.Edges.Flags, func(flag *ent.MessageFlag) string {
-			return flag.Value
-		}))
+			result = append(result, mfs)
+		}
```
jameshoulahan, line 350: the `AddSelf` operations here would probably do a lot of unnecessary appends and reallocations; can you collect the msg.Edges.Flags values with one `xslices.Map` call and `AddSelf(flags...)` in one go?

LBeernaertProton: I'm avoiding the the allocation of a temporary slice here, since map returns a new slice.

jameshoulahan (dismissed):

jameshoulahan (approved):

PR #155, opened 2022-09-27
feat: Allow connector to persist initial UID validity of mailboxes

jameshoulahan:


internal/db/mailbox.go
```diff
@@ -70,27 +70,41 @@ func RenameMailboxWithRemoteID(ctx context.Context, tx *ent.Tx, mboxID imap.Labe
...
 		Exec(ctx); err != nil {
-		return err
+		return 0, false, err
 	}
 
-	return nil
+	return newUIDValidity, newUIDValidity > curUIDValidity, nil
 }
 
 func UpdateRemoteMailboxID(ctx context.Context, tx *ent.Tx, internalID imap.InternalMailboxID, remoteID imap.LabelID) error {
```
LBeernaertProton, line 110: Is this function unused now?

internal/db/mailbox.go
```diff
@@ -70,27 +70,41 @@ func RenameMailboxWithRemoteID(ctx context.Context, tx *ent.Tx, mboxID imap.Labe
 	return nil
 }
 
-func DeleteMailboxWithRemoteID(ctx context.Context, tx *ent.Tx, mboxID imap.LabelID) error {
-	// get uid validity
+func DeleteMailboxWithRemoteID(ctx context.Context, tx *ent.Tx, mboxID imap.LabelID) (imap.UID, bool, error) {
```
LBeernaertProton, line 73: Can you document what the bool indicates?

internal/db/mailbox.go
```diff
@@ -303,42 +317,33 @@ func FilterMailboxContains(ctx context.Context, client *ent.Client, mboxID imap.
...
-		Only(ctx)
-
-	if ent.IsNotFound(err) {
-		_, err := tx.UIDValidity.Create().
-			SetUIDValidity(mailbox.DefaultUIDValidity).
-			Save(ctx)
-		if err != nil {
-			return 0, err
+func InitGlobalUIDValidity(ctx context.Context, tx *ent.Tx, uidValidity imap.UID) error {
+	if _, err := tx.UIDValidity.Query().Select(uidvalidity.FieldUIDValidity).Only(ctx); err != nil {
```
LBeernaertProton, line 321: Should we maybe update the UID validity here if the value reported by the connector is higher than the one we have now?

internal/db/mailbox.go
jameshoulahan: It is still used and is not related to these changes. Are you referring to a different one?

internal/db/mailbox.go
jameshoulahan: done

internal/db/mailbox.go
LBeernaertProton: Sorry, my mistake.

internal/db/mailbox.go
jameshoulahan: I don't think so; the validity reported by the connector should never be higher than what we have, since we're the only ones ones who increase it. If it is higher, something's gone wrong 🤔

LBeernaertProton: What if the user is using an outdated version of the cache that they have left lying around and decide to plug it in now?

LBeernaertProton: Yes, but in our current state you can still do this and then have the remote be set to a lower UID validity than what the connector may already have.

LBeernaertProton:
I just noticed that since we get the UIDValidity from the connector, update the connector, we might as well just keep it in memory and skip the database all together :thinking:

internal/db/mailbox.go
jameshoulahan: Good point, we should at least be safe regarding that

jameshoulahan: Done

LBeernaertProton (approved):

PR #156, opened 2022-09-27
feat(GODT-1609):use bytes array instead of string for password

rlejeune74:


connector/dummy.go
```diff
@@ -82,8 +83,8 @@ func NewDummy(usernames []string, password string, period time.Duration, flags,
 	return conn
 }
 
-func (conn *Dummy) Authorize(username, password string) bool {
-	if password != conn.password {
+func (conn *Dummy) Authorize(username string, password []byte) bool {
+	if !reflect.DeepEqual(password, conn.password) {
```
jameshoulahan, line 87: reflect is slow, for this you should use `bytes.Equal` (or `bytes.Compare` if you want to avoid converting back to strings)

tests/run_test.go
```diff
@@ -21,7 +21,7 @@ func runOneToOneTest(tb testing.TB, options *serverOptions, tests func(*testConn
 func runOneToOneTestWithAuth(tb testing.TB, options *serverOptions, tests func(*testConnection, *testSession)) {
 	runOneToOneTest(tb, options, func(c *testConnection, s *testSession) {
 		withTag(func(tag string) {
-			c.Cf("%v login %v %v", tag, options.defaultUsername(), options.defaultUserPassword()).OK(tag)
+			c.Cf("%v login %v %v", tag, options.defaultUsername(), string(options.defaultUserPassword())).OK(tag)
```
jameshoulahan, line 24: you can use `%s` instead of `%v` to print bytes as a string

tests/run_test.go
```diff
@@ -21,7 +21,7 @@ func runOneToOneTest(tb testing.TB, options *serverOptions, tests func(*testConn
 func runOneToOneTestWithAuth(tb testing.TB, options *serverOptions, tests func(*testConnection, *testSession)) {
 	runOneToOneTest(tb, options, func(c *testConnection, s *testSession) {
 		withTag(func(tag string) {
-			c.Cf("%v login %v %v", tag, options.defaultUsername(), options.defaultUserPassword()).OK(tag)
+			c.Cf("%v login %v %s", tag, options.defaultUsername(), string(options.defaultUserPassword())).OK(tag)
```
jameshoulahan, line 24: my point was, if you use `%s`, you don't need to convert bytes to string before printing :p

```suggestion
			c.Cf("%v login %v %s", tag, options.defaultUsername(), options.defaultUserPassword()).OK(tag)
```

rlejeune74: :facepalm:

PR #158, opened 2022-09-28
Improve search and add -imap-search-cmd benchmark

LBeernaertProton:


internal/db/message.go
```diff
@@ -236,6 +236,10 @@ func GetMessage(ctx context.Context, client *ent.Client, messageID imap.Internal
 	return client.Message.Query().Where(message.ID(messageID)).Only(ctx)
 }
 
+func GetMessageDataAndSize(ctx context.Context, client *ent.Client, messageID imap.InternalMessageID) (*ent.Message, error) {
```
jameshoulahan, line 239: `Data` --> `Date`?

internal/state/mailbox_search.go
```diff
@@ -4,425 +4,581 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"github.com/ProtonMail/gluon/internal/db"
+	"github.com/bradenaw/juniper/parallel"
+	"github.com/bradenaw/juniper/xslices"
 	"net/mail"
+	"runtime"
 	"strings"
+	"sync/atomic"
```
jameshoulahan, line 13: might wanna `goimports` this, shouldn't be mixing stdlib and extern imports

internal/state/mailbox_search.go
```diff
@@ -4,425 +4,581 @@ import (
...
-		if contexts.IsUID(ctx) {
-			return uint32(msg.UID)
+	msgCount := m.snap.len()
+
+	result := make([]uint32, msgCount)
+
+	activeSearchRequests := atomic.AddInt32(&totalActiveSearchRequests, 1)
+	defer atomic.AddInt32(&totalActiveSearchRequests, -1)
+
+	parallelism := runtime.NumCPU() / int(activeSearchRequests)
```
jameshoulahan, line 51: we do the same thing for fetch -- wondering two things:
- should we merge the two counters? i'd expect fetching and searching to take each other into account wrt parallelism
- should we abstract this logic into an object to more tidily handle this? the object could return you the amount of parallelism you're allowed to use at any given time

jameshoulahan:
```go
type limiter struct {
  // hold the number of ongoing jobs
}

func (l *limiter) do(fn func(limit int)) {
  // determine how much parallelism can be applied right now
  
  // increase ongoing job counter
  
  fn(limit) // call the callback with the amount allowed
  
  // decrease ongoing job counter
}
```

internal/state/mailbox_search.go
```diff
@@ -4,425 +4,581 @@ import (
...
+			return searchData{}, err
 		}
 
-		candidates = filtered
+		data.literal = l
 	}
 
-	return candidates, nil
+	if op.needsHeader {
+		headerBytes, _ := rfc822.Split(data.literal)
```
jameshoulahan, line 103: could probably optimize this with a separate `rfc822.SplitHeader`  and `rfc822.SplitBody` that doesn't unnecessarily return the header/body?

internal/state/mailbox_search.go
LBeernaertProton: I initially thought about limiting the total operations for both fetch and search. But I found that since most of these queries are IO bound, they don't really stress the CPU a lot. I never go 100% cpu usage during these benchmarks. 

I think it's safe to have ncpu fetches and search in parallel.

I would leave this query decision to each command. E.g: with search it makes sense to parallelize all the time since you need to scan the whole inbox.

internal/state/mailbox_search.go
LBeernaertProton: It's returning sub-slices, no copies are performed.

internal/state/mailbox_search.go
```diff
@@ -4,425 +4,581 @@ import (
...
+
+		data.header = h
+	}
+
+	return data, nil
 }
 
-func (m *Mailbox) matchSearchKey(ctx context.Context, candidates []snapMsgWithSeq, key *proto.SearchKey, decoder *encoding.Decoder) ([]snapMsgWithSeq, error) {
+func applySearch(ctx context.Context, m *Mailbox, msg snapMsgWithSeq, searchOp buildSearchOpResult) (bool, error) {
+	data, err := buildSearchData(ctx, m, &searchOp, msg)
```
jameshoulahan, line 117: if we're only going to pass around pointers to the search op (and the data below), it's probably better to just construct them on the heap in the first place, giving us smaller stack sizes, no?

```go
func newBuildSearchOpResult(op searchOp, needs ...searchOpResultOption) *buildSearchOpResult {
	r := &buildSearchOpResult{op: op}

	...
}
```

```go
func buildSearchData(ctx context.Context, m *Mailbox, op *buildSearchOpResult, message snapMsgWithSeq) (*searchData, error) {
	data := &searchData{message: message}
...
```

```

internal/state/mailbox_search.go
LBeernaertProton: Done

internal/state/mailbox_search.go
jameshoulahan: hmm okay 👍

internal/state/mailbox_search.go
LBeernaertProton: Done

jameshoulahan (approved):

PR #160, opened 2022-09-29
fix: Don't close connector

jameshoulahan:
This change stops the gluon backend from closing its connectors. It shouldn't be gluon's responsibility to close the connectors; rather, they should be closed by the surrounding parent code that passes them in to gluon. This enables, for example, a single connector to survive across gluon restarts.

### DIFF
--- a/benchmarks/gluon_bench/utils/connector_factory.go
+++ b/benchmarks/gluon_bench/utils/connector_factory.go
@@ -13,6 +13,7 @@ import (
 type ConnectorImpl interface {
 	Connector() connector.Connector
 	Sync(ctx context.Context) error
+	Close()
 }
 
 type ConnectorBuilder interface {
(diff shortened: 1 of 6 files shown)


LBeernaertProton (commented):
I'm a bit concerned about the what happens to lingering references to the update channel of the connector.

We can now potentially have issues with the connector being shared by two separate user if used incorrectly and/or old user instance applying updates to intended for the new user.

If the sharing of connector state is something that you wish to preserve, I would suggest you implement this as part of  connector implementation rather than at Gluon the gluon level. Unless there's a motivation here that's not influenced by Bridge.

LBeernaertProton:
Please apply the changes to Demo and Benchmark projects as well.

jameshoulahan:
Closing, decided to change the implementation in bridge so as to not require these changes on the gluon side

PR #168, opened 2022-10-03
feat: Back to transaction-free on-disk store

jameshoulahan:


LBeernaertProton (commented):
We should add a cleanup routine on startup that deletes all cached files on disk since it's possible to now have stale data in the cache that will never be removed. E.g.: Add/delete rollback.

PR #170, opened 2022-10-05
feat(GODT-1891): kick client that send too many invalid IMAP command

rlejeune74:


LBeernaertProton (changes requested):

internal/session/session.go
```diff
@@ -165,10 +171,18 @@ func (s *Session) serve(ctx context.Context) error {
 
 			if res.err != nil {
 				logrus.WithError(res.err).Debugf("Error during command parsing")
+				s.errorCount++
 
 				if errors.Is(res.err, io.EOF) {
 					logrus.Debugf("Connection to client lost")
 					return nil
+				} else if s.errorCount >= maxSessionError {
```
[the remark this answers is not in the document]
rlejeune74: Indeed I've checked on Dovecot they are resetting the counter on good command.

internal/session/errors.go
```diff
@@ -11,5 +11,7 @@ var (
 	ErrNotAuthenticated     = errors.New("session is not authenticated")
 	ErrAlreadyAuthenticated = errors.New("session is already authenticated")
 
+	ErrTooManyInvalid = errors.New("too many invalid IMAP command")
```
jameshoulahan, line 14:
```suggestion
	ErrTooManyInvalid = errors.New("too many invalid IMAP commands")
```

tests/session_test.go
```diff
@@ -324,3 +324,35 @@ func parseDateFromDelimiter(messageDelimiter string) (t time.Time, err error) {
...
+	})
+}
+
+func TestResetTooManyInvalidCommand(t *testing.T) {
+	runOneToOneTestWithAuth(t, defaultServerOptions(t), func(c *testConnection, s *testSession) {
+		for i := 1; i <= 19; i++ {
+			c.C(fmt.Sprintf("%d FOO", i))
+			c.Sx(fmt.Sprintf("%d BAD mismatched input", i))
+		}
+		c.C(fmt.Sprintf("100 NOOP"))
```
jameshoulahan, line 347:
```suggestion
		c.C("100 NOOP")
```

tests/session_test.go
```diff
@@ -324,3 +324,35 @@ func parseDateFromDelimiter(messageDelimiter string) (t time.Time, err error) {
...
+		c.Sx(fmt.Sprintf("100 OK"))
```
jameshoulahan, line 348:
```suggestion
		c.C("100 NOOP").OK("100")
```

no need for `fmt.Sprintf`, and you can also just expect an OK like this

tests/session_test.go
```diff
@@ -324,3 +324,35 @@ func parseDateFromDelimiter(messageDelimiter string) (t time.Time, err error) {
...
+		for i := 1; i <= 19; i++ {
+			c.C(fmt.Sprintf("%d FOO", i))
```
jameshoulahan, line 350:
```suggestion
			c.Cf("%d FOO", i)
```

tests/session_test.go
```diff
@@ -324,3 +324,35 @@ func parseDateFromDelimiter(messageDelimiter string) (t time.Time, err error) {
...
+			c.Sx(fmt.Sprintf("%d BAD mismatched input", i))
+		}
+		c.C(fmt.Sprintf("100 FOO"))
+		c.S(fmt.Sprintf("100 BAD too many invalid IMAP command"))
```
jameshoulahan:
```suggestion
		c.C("100 FOO").BAD("100")
```
you can do this too (if you don't care about the exact error string, which maybe you do)

LBeernaertProton (approved):

PR #171, opened 2022-10-05
feat(GODT-1643): RFC5322 Parser in C++

LBeernaertProton:
Rewrite of https://github.com/ProtonMail/go-rfc5322 using C++.

We also add the `d` postfix to debug builds to avoid having debug binaries linked into go.


imap/envelope.go
```diff
@@ -76,9 +78,14 @@ func envelope(header *rfc822.Header, c *paramList, writer parListWriter) error {
 
 // TODO: Should use RFC5322 package here but it's too slow... sad.
 func tryParseAddressList(val string) []*mail.Address {
-	if addr, err := mail.ParseAddressList(val); err == nil {
+	/*if addr, err := mail.ParseAddressList(val); err == nil {
 		return addr
+	}*/
```
jameshoulahan, line 83: can remove this now

imap/envelope.go
```diff
@@ -76,9 +78,14 @@ func envelope(header *rfc822.Header, c *paramList, writer parListWriter) error {
 
 // TODO: Should use RFC5322 package here but it's too slow... sad.
```
jameshoulahan, line 79: and can remove this too

jameshoulahan:
I'd like you to also include the XML test set from the go-rfc5322 package here. They were very valuable. Should be simple enough to test them from the go side.

internal/parser/rfc5322_address_list.go
```diff
@@ -0,0 +1,39 @@
...
+	"net/mail"
+	"unsafe"
+)
+
+func ParseRFC5322AddressList(input string) ([]*mail.Address, error) {
+	parser := C.RFC5322AddressList_new()
+	defer C.RFC5322AddressList_free(parser)
+
+	cstr := C.CString(input)
+	defer C.free(unsafe.Pointer(cstr))
```
jameshoulahan, line 20: i like how we avoided swig here, do you think that's something we could realistically do with the IMAP parser too? i used swig because the IMAP parser expects C++ types as input, but i guess we could rewrite them into pure C types

LBeernaertProton: I will do a pass over the IMAP parser once I'm done with this and I'll see what I can do from there.

jameshoulahan (approved):

PR #174, opened 2022-10-06
feat(GODT-1587):Add a wait time after too many failed login attempt

rlejeune74:


internal/session/handle_login.go
```diff
@@ -24,9 +31,21 @@ func (s *Session) handleLogin(ctx context.Context, tag string, cmd *proto.Login,
 
 	state, err := s.backend.GetState(cmd.GetUsername(), cmd.GetPassword(), s.sessionID)
 	if err != nil {
+		s.loginErrorCount++
```
LBeernaertProton, line 34: Wouldn't it be better to have this check on the the user account rather than the session? An evil user could just launch N connection in parallel. If you have this check on the user, you can scale the login timeout based on the number of attempts. 

One should still disconnect session after 3 tries, but all login attempts should have the timeout scaling applied.

internal/backend/backend.go
```diff
@@ -28,6 +31,10 @@ type Backend struct {
 
 	// storeBuilder builds stores for the backend users.
 	storeBuilder store.Builder
+
+	/// loginErrorCount login failure counter
+	loginErrorCount int
```
LBeernaertProton, line 36: Can you use sync atomics rather than a mutex here?

LBeernaertProton (changes requested):

internal/backend/backend.go
```diff
@@ -167,13 +179,41 @@ func (b *Backend) Close(ctx context.Context) error {
 	return nil
 }
 
-func (b *Backend) getUserID(username string, password []byte) (string, error) {
+func (b *Backend) getUserID(ctx context.Context, username string, password []byte) (string, error) {
+	if atomic.LoadInt32(&b.loginErrorCount) >= maxLoginAttempts {
+		b.lockLogin.Lock()
```
LBeernaertProton, line 184: When using atomics, you don't need to use the mutex.

internal/backend/backend.go
```diff
@@ -167,13 +179,41 @@ func (b *Backend) Close(ctx context.Context) error {
...
 		if user.connector.Authorize(username, password) {
+			atomic.StoreInt32(&b.loginErrorCount, 0)
 			return user.userID, nil
 		}
 	}
 
+	atomic.AddInt32(&b.loginErrorCount, 1)
+
+	if atomic.LoadInt32(&b.loginErrorCount) == maxLoginAttempts {
+		b.notifyLoginBlocked(true)
```
LBeernaertProton, line 199: The connector should internally handle this. Keep in mind that one connector could potentially authorize a lot of different users. In this case you are blocking the login for all the users rather than the one where the authentication actually failed.

internal/backend/backend.go
rlejeune74: the lock is here to delayed the answer so that the IMAP client wont see that the login is blocked but rather the answer will take longer. (as Jakub mentionned)
here the purpose is really to wait for the timer to end before treating a new login attempt.

internal/backend/backend.go
rlejeune74: On login fails we don't know what is the user that failed to login, do we ?
Idea here was to notify any connector that we locked the Login procedure for a while and then the Connector can decide what we want to do with that.
For simple implementation I was just thinking of printing text at login page. Rather than a toggle off mechanism that is a bit more advanced.

internal/backend/backend.go
rlejeune74: done

internal/session/handle_login.go
rlejeune74: we can't do it on user account  side since we also need to track bad login attempts with unknown account/user

rlejeune74: I did modification to handle the timeout at a more global level (backend) rather that only on sessions.

internal/backend/backend.go
LBeernaertProton: I see. In that case there is still a small chance that two clients login in at the same time with the invalid credentials and one of them will reset the login count before the timer has a chance to start.

E.g.:

Client A  fails login and increases the login count to `maxLoginAttempts` and before it has a chance top reach the go routine that launches the timer client B also fails the login and acquires the mutex before the goroutine can. Now B resets the counter before the timer of A runs to completion.

internal/backend/backend.go
rlejeune74: but still doable (toggle off mechanism from connector)

rlejeune74: I understand the connector/user oriented part if the login attempt  target a known user, but how can we handle login attempts that are not targeting known user ?

I also understand that by doing global tempo/lock, we can easily deny the Login mechanism for all connector by spamming wrong Login.
Also only handling known user could provide a hint while trying to force the login since we can identify by the locking mechanism that the user is known. (but at least still better than denying the login process).

If we go that way, there is nothing to do within Gluon and everything should be handle on connector side with the Authorize method (that is still called for every connector until one says OK).
Is that what you have in mind ?

LBeernaertProton: I was trying to say we could keep the current logic but without the extra connector method.

 If we really wanted to implement more advanced error pattern matching we need to resort to IP filtering. An interface to implement the behavior would be the better option. That way once could tailor the behavior for Bridge and any other users of the library.

The block all equivalent could be exactly the code you have submitted here.

internal/backend/backend.go
LBeernaertProton: The proposed changes address the counter reset issue, but don't exactly force the client to wait until the timer has finished in the example listed above.

rlejeune74: True... always have to figure out how to handle this

LBeernaertProton (approved):

PR #176, opened 2022-10-10
fix(GODT-1570): Handle literals in IMAP Parser without map + C API

LBeernaertProton:
Adjust the parser so that we don't have to pass in a literal map to parse the message. We removed the Swig API in favor of a simpler C style API and commands are no longer serialized to string but straight to a byte array.

This patch also changes the representation of the literal as it is supposed to be a single character rather than a string.


internal/liner/liner.go
```diff
@@ -24,13 +23,15 @@ func New(r io.Reader) *Liner {
 // Read reads a full line, automatically reading again if the line was not complete.
 // Each time an additional read is performed, doContinuation is called.
 // If the callback returns an error, the operation is aborted.
-func (l *Liner) Read(doContinuation func() error) ([]byte, map[string][]byte, error) {
+func (l *Liner) Read(doContinuation func() error) (string, error) {
```
cuthix, line 26: wouldn't be better to return `[]byte` ?

tests/server_test.go
```diff
@@ -109,7 +109,7 @@ func withIdleBulkTime(idleBulkTime time.Duration) serverOption {
 }
 
 func withDelimiter(delimiter string) serverOption {
```
cuthix, line 111: I am not sure what was the original motivation for `string -> rune` but I think we should keep it consistent

```suggestion
func withDelimiter(delimiter rune) serverOption {
```

internal/parser/grammar/IMAPParser.g4
```diff
@@ -4,8 +4,12 @@ options {
 	tokenVocab = IMAPLexer;
 }
 
+@members {
+    virtual std::string consumeTokens(std::string) = 0;
+}
+
 // 2.2. Commands and Responses
-crlf: CR LF;
+crlf: CR? LF;
```
jameshoulahan, line 12: :scream: why? shouldn't CRLF _always_ be CR and LF together?

internal/parser/grammar/IMAPParser.g4
```diff
@@ -36,7 +40,9 @@ quotedChar
 
 quotedSpecial: DQuote | Backslash;
 
-literal: LCurly number RCurly crlf uuid;
+literal locals[
+std::string mLiteralText
+]: LCurly number{auto numberCtx=$number.ctx;}RCurly crlf {$mLiteralText = $parser->consumeTokens(numberCtx->getText());};
```
jameshoulahan, line 45: mmm okay i was wondering how you did this, and i see you used injected code -- i guess at this point we've accepted being tied to C++, i think this is a good choice :+1:

internal/liner/liner.go
LBeernaertProton: we needed to convert to string for the C++ parser. This type has no other use other than to prepare the input for the parser.

internal/parser/grammar/IMAPParser.g4
LBeernaertProton: Woops, forgot to revert this :sweat_smile:

internal/parser/src/imap/parser.h
```diff
@@ -14,12 +15,14 @@ struct ParseResult {
   // If not empty, contains the currently parsed command tag
   std::string tag;
   // Contains the protobuf serialized data
-  std::string command;
+  std::vector<uint8_t> command;
```
jameshoulahan, line 18: can we align between `uint8_t` and `char` here and the methods below? or do we need to stick with `char` below for the C API?

tests/connection_test.go
```diff
@@ -197,14 +196,10 @@ func (s *testConnection) doBench(b *testing.B, cmd string) {
 
 // TODO: This is shitty because the uuid of one literal may appear within the data of another literal.
```
jameshoulahan, line 197: i think you can remove this TODO now

tests/server_test.go
jameshoulahan, line 111: option could take `rune` as well?

LBeernaertProton:
I"m afraid I jumped the gun too soon. I failed to realize the the skip code I added was operating on tokens rather than bytes.

PR #177, opened 2022-10-10
feat: Give update.Wait() a context parameter so waiting can be cancelled

jameshoulahan:


imap/update_waiter.go
```diff
@@ -23,6 +32,17 @@ func (w *updateWaiter) Wait() {
 	w.wg.Wait()
 }
 
+func (w *updateWaiter) WaitContext(ctx context.Context) {
+	waitCh := make(chan struct{})
+
```
LBeernaertProton, line 37: Random thought. Should we maybe change this type to just use a channel instead of of a sync.WG?

jameshoulahan: maybe, but i think i like the waitgroup here, it's safer because a default constructed waitgroup is usable but a default constructed channel isn't, and the wait/done methods map more logically those of a waitgroup

LBeernaertProton (approved):

PR #181, opened 2022-10-11
test: Reduce login jail time in tests

jameshoulahan:


LBeernaertProton (commented):
I think this would be better if we could configure the login time out with a gluon server config option.

jameshoulahan:
> I think this would be better if we could configure the login time out with a gluon server config option.

Agreed, done

LBeernaertProton (dismissed):

PR #187, opened 2022-10-14
feat(GODT-1775): do not log private information nor raw data

rlejeune74:


rlejeune74:
This is only handling IMAP Output filtering.
Input filtering needs more work since it requires then to add an additional parsing process to apply the filter.

PR #190, opened 2022-10-19
feat: Expose ParseMediaType

jameshoulahan:
The standard library mime.ParseMediaType does not handle some edge cases. This commit exposes a package-level ParseMediaType function, allowing the standard library calls to be replaced with a custom function.

### DIFF
--- a/imap/structure.go
+++ b/imap/structure.go
@@ -2,7 +2,6 @@ package imap
 
 import (
 	"bytes"
-	"mime"
 	"strings"
 
 	"github.com/ProtonMail/gluon/rfc822"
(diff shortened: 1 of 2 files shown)


rfc822/mime.go
```diff
@@ -5,6 +5,9 @@ import (
 	"strings"
 )
 
+// ParseMediaType parses a MIME media type.
+var ParseMediaType = mime.ParseMediaType
```
LBeernaertProton, line 9: Why does this need to be a global variable?

imap/structure.go
```diff
@@ -147,7 +146,7 @@ func getMIMEInfo(section *rfc822.Section) (string, string, map[string]string, er
 }
 
 func addDispInfo(c *paramList, writer parListWriter, header *rfc822.Header) {
-	if contentDisp, contentDispParams, err := mime.ParseMediaType(header.Get("Content-Disposition")); err == nil {
+	if contentDisp, contentDispParams, err := rfc822.ParseMediaType(header.Get("Content-Disposition")); err == nil {
```
LBeernaertProton, line 149: Should we remove the old `ParseMediaType` altogether now?

jameshoulahan: The old one is provided by standard lib, there's nothing to remove

rfc822/mime.go
LBeernaertProton: I'm not a big fan of this global variable to replace functionality. I Would rather have this as an input to the actual parsing function.

jameshoulahan: That would require a much more significant change spanning multiple modules (here, bridge, liteapi). Doing it this way is quite common; we already do it in go-rfc5322, and it's done by other go libs as well (e.g. [here](https://github.com/emersion/go-message/blob/master/charset.go#L35))

LBeernaertProton (approved):

PR #191, opened 2022-10-19
fix(GODT-1760): Fix out of order UID panics

LBeernaertProton:
Extendend snapshot with `appendMessageFromOtherState` to enable out of order UID insertion. Please check the comments in
`TestAppendCanHandleOutOfOrderUIDUpdates` for why this is necessary.

### DIFF
--- a/tests/append_test.go
+++ b/tests/append_test.go
@@ -3,6 +3,7 @@ package tests
 import (
 	"bytes"
 	"context"
+	"fmt"
 	"github.com/ProtonMail/gluon/connector"
 	"github.com/ProtonMail/gluon/imap"
 	"os"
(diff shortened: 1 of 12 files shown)


tests/append_test.go
```diff
@@ -261,3 +261,42 @@ func TestAppendConnectorReturnsSameRemoteIDDifferentMBox(t *testing.T) {
...
+		wg.Add(2)
+
+		for i := 0; i < 2; i++ {
+			go func(index int) {
+				defer wg.Done()
+				appendFN(index)
+			}(i)
+		}
+
+		wg.Wait()
```
jameshoulahan, line 300: at this point, after clients have received all updates from each other, could we perform a list with each client and assert that they all see the same messages in the same order?

tests/append_test.go
```diff
@@ -261,3 +261,42 @@ func TestAppendConnectorReturnsSameRemoteIDDifferentMBox(t *testing.T) {
...
+	// Make sure we are correctly handling the case where we have to clients doing append at the same time.
+	// Both clients append a message and get assigned UID according to whichever got there first:
+	//
+	// * Client A -> Append -> UID 1
+	// * Client B -> Append -> UID 2
+	//
+	// All Clients apply their changes to their local state immediately and will receive a deferred updates for the
+	// same mailbox if other clients make updates.
+	// In the case of client B, it appends UID2 as the first message and then later receives an update from A with
+	// an UID lower than the last UID which causes a panic.
```
cuthix, line 275:
```suggestion
	// an UID lower than the last UID caused unnecessary panic in past.
```

LBeernaertProton: Done

tests/append_test.go
LBeernaertProton: Good point. Fixed. I also had to add option to disable parallelism to make sure the fetch results arrive in order.

PR #193, opened 2022-10-20
Chore: Rename Label to Mailbox in connector interface

LBeernaertProton:


connector/dummy.go
```diff
@@ -36,8 +36,8 @@ type Dummy struct {
 	// These hold the default flags/attributes given to mailboxes.
 	flags, permFlags, attrs imap.FlagSet
 
-	// These hold prefixes given to folders (exclusive) and labels (non-exclusive).
-	pfxFolder, pfxLabel string
+	// These hold prefixes given to folders (exclusive) and mailboxes (non-exclusive).
```
jameshoulahan, line 39: okay here i think it's best not to rename; this naming is very confusing :D i think it's okay for the dummy to keep proton-specific stuff, as it's implementing a proton-specific version of the connector

LBeernaertProton: Done

connector/dummy.go
```diff
@@ -36,7 +36,7 @@ type Dummy struct {
 	// These hold the default flags/attributes given to mailboxes.
 	flags, permFlags, attrs imap.FlagSet
 
-	// These hold prefixes given to folders (exclusive) and labels (non-exclusive).
+	// These hold prefixes given to folders (exclusive) and mailboxes (non-exclusive).
```
jameshoulahan, line 39:
```suggestion
	// These hold prefixes given to folders (exclusive) and labels (non-exclusive).
```

LBeernaertProton: Done

jameshoulahan (approved):

PR #196, opened 2022-10-24
Add Section.Walk and Section.DecodedBody

jameshoulahan:


LBeernaertProton (commented):
Looks good, but can you add some tests to make sure it's working as intended?

jameshoulahan:
> Looks good, but can you add some tests to make sure it's working as intended?

Done

LBeernaertProton (approved):

PR #202, opened 2022-10-24
feat: Improve goroutine annotation

jameshoulahan:
This change adds a logging package which makes it easy to annotate goroutines with labels that appear in pprof profiles.

Existing annotations were converted to use the new package.

### DIFF
--- a/internal/session/handle.go
+++ b/internal/session/handle.go
@@ -3,12 +3,11 @@ package session
 import (
 	"context"
 	"fmt"
-	"runtime/pprof"
-	"strconv"
 
 	"github.com/ProtonMail/gluon/internal/parser/proto"
 	"github.com/ProtonMail/gluon/internal/response"
 	"github.com/ProtonMail/gluon/internal/state"
+	"github.com/ProtonMail/gluon/logging"
 	"github.com/ProtonMail/gluon/profiling"
 	"github.com/ProtonMail/gluon/queue"
 )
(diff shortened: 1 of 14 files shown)


internal/session/handle.go
```diff
@@ -32,19 +31,20 @@ func (s *Session) handleOther(
...
 				if res, ok := response.FromError(err); ok {
 					resCh <- res
 				} else {
 					resCh <- response.No(tag).WithError(err)
 				}
 			}
+		}, logging.Labels{
+			"Action":    "Handling IMAP command",
+			"SessionID": s.sessionID,
+			"Command":   cmd,
```
LBeernaertProton, line 47: Potential data/security issue here. Maybe we should just try to get the name of the command rather than the full object?

logging/logging.go
```diff
@@ -0,0 +1,48 @@
...
+	FnKey = "fn"
+
+	// FileKey refers to the file name.
+	FileKey = "file"
+
+	// LineKey refers to the line number.
+	LineKey = "line"
+)
+
+func GoAnnotated(ctx context.Context, fn func(context.Context), labelMap ...Labels) {
```
LBeernaertProton, line 23: Since we have this nicely encapsulated, do you think you could add a build mode where all the pprof stuff is not included?

LBeernaertProton (changes requested):

internal/session/handle.go
jameshoulahan: Oops yes i actually intended to just get the commands name but couldn't figure it out so left this as a placeholder but forgot about it. Good catch. Will see if i can get the command name, otherwise will remove

logging/logging.go
jameshoulahan: Will do, yes

internal/session/handle.go
jameshoulahan: Oops yes i actually intended to just get the commands name but couldn't figure it out so left this as a placeholder but forgot about it. Good catch. Will see if i can get the command name, otherwise will remove

logging/logging.go
jameshoulahan: done, annotation is disabled if `gluon_pprof_disabled` is set

internal/session/handle.go
jameshoulahan: Tried to get command name but go doesn't make it easy, so removed this line. If need be, can add back later with proper fix.

LBeernaertProton (approved):

PR #208, opened 2022-10-26
fix(GODT-1913): Draft: add panic handler.

cuthix:


server.go
```diff
@@ -316,6 +323,8 @@ func (s *Server) newEventCh(ctx context.Context) chan events.Event {
 	eventCh := make(chan events.Event)
 
 	logging.GoAnnotated(ctx, func(ctx context.Context) {
+		defer s.panicHandler.HandlePanic()
```
LBeernaertProton, line 326: Is this supposed to be added to every go routine?

cuthix: I am not sure how much we need to cover

PR #209, opened 2022-10-27
fix: Fix state connector deadlock

LBeernaertProton:
Remove `refresh` function as it is no longer required since we removed the operation queue. Additionally this function was causing a deadlock on error since the update it was waiting on could not be executed as the current operation has acquired the database lock.

### DIFF
--- a/internal/backend/state_connector_impl.go
+++ b/internal/backend/state_connector_impl.go
@@ -46,7 +46,7 @@ func (sc *stateConnectorImpl) CreateMailbox(ctx context.Context, name []string) 
 	return mbox, nil
 }
 
-func (sc *stateConnectorImpl) UpdateMailbox(ctx context.Context, mboxID imap.MailboxID, oldName, newName []string) error {
+func (sc *stateConnectorImpl) UpdateMailbox(ctx context.Context, mboxID imap.MailboxID, newName []string) error {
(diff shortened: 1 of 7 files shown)


internal/backend/state_connector_impl.go
```diff
@@ -165,33 +145,3 @@ func (sc *stateConnectorImpl) newContextWithMetadata(ctx context.Context) contex
 
 	return ctx
 }
-
-func (sc *stateConnectorImpl) refresh(ctx context.Context, messageIDs []imap.MessageID, mboxIDs ...imap.MailboxID) error {
-	for _, messageID := range messageIDs {
-		message, mboxIDs, err := sc.connector.GetMessage(ctx, messageID)
```
jameshoulahan, line 171: could you please also remove `GetMessage` and `GetMailbox` from `connector.Connector`? Those were only needed for refresh

jameshoulahan: They're dangerous, i only reluctantly implemented them before, because they allow gluon to peek at server state before it's ready -- every update should come through the update channel, not directly from GetMessage/GetLabel

LBeernaertProton: Done

jameshoulahan (approved):

PR #211, opened 2022-10-28
fix: Don't set the Internal ID in the header twice

jameshoulahan:
We set it above, so don't do it again when adding to the store.

### DIFF
--- a/internal/backend/connector_updates.go
+++ b/internal/backend/connector_updates.go
@@ -236,12 +236,7 @@ func (user *user) applyMessagesCreated(ctx context.Context, update *imap.Message
 		if err := user.db.Write(ctx, func(ctx context.Context, tx *ent.Tx) error {
 			// Create messages in the store
 			for _, msg := range chunk {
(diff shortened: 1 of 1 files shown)


internal/backend/connector_updates.go
```diff
@@ -227,11 +227,6 @@ func (user *user) applyMessagesCreated(ctx context.Context, update *imap.Message
 		if err := user.db.Write(ctx, func(ctx context.Context, tx *ent.Tx) error {
 			// Create messages in the store
 			for _, msg := range chunk {
-				literalWithHeader, err := rfc822.SetHeaderValue(msg.Literal, ids.InternalIDKey, msg.InternalID.String())
-				if err != nil {
-					return fmt.Errorf("failed to set internal ID: %w", err)
-				}
-
 				if err := user.store.Set(msg.InternalID, literalWithHeader); err != nil {
```
LBeernaertProton, line 230: Shouldn't this be msg.Literal?

jameshoulahan: :facepalm: too many things happening at once, fixed

LBeernaertProton (approved):

PR #216, opened 2022-11-03
feat: MessageUpdated Update

LBeernaertProton:
To be used in cases where one needs to atomically replace a message (e.g: Drafts) in one go. This was previously possible with a combination of Delete + Create update, but these were not guaranteed to be atomic and could cause undefined behaviors if something else happened in between.

The old logic to handle the replacement of delete messages on create has been reverted. It is now expected that this type of behavior uses the newly introduced update.

### DIFF
--- a/imap/update_message_updated.go
+++ b/imap/update_message_updated.go
@@ -1,0 +1,38 @@
+package imap
+
+import (
+	"fmt"
+	"github.com/bradenaw/juniper/xslices"
+)
+
+// MessageUpdated replaces the previous behavior of MessageDelete followed by MessageCreate. Furthermore, it guarantees
+// that the operation is executed atomically.
+type MessageUpdated struct {
+	updateBase
+	*updateWaiter
+
+	Message       Message
+	Literal       []byte
+	MailboxIDs    []MailboxID
+	ParsedMessage *ParsedMessage
+}
+
+func NewMessageUpdated(message Message, literal []byte, mailboxIDs []MailboxID, parsedMessage *ParsedMessage) *MessageUpdated {
+	return &MessageUpdated{
+		updateWaiter:  newUpdateWaiter(),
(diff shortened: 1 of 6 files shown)


imap/update_message_updated.go
```diff
@@ -0,0 +1,38 @@
+package imap
+
+import (
+	"fmt"
+	"github.com/bradenaw/juniper/xslices"
+)
+
+// MessageUpdated behaves the same as a MessageDelete followed by MessageCreate. It just guarantees that the both ot
```
jameshoulahan, line 8:
```suggestion
// MessageUpdated behaves the same as a MessageDelete followed by MessageCreate. It just guarantees that the both of
```

LBeernaertProton: Done

internal/backend/connector_updates.go
```diff
@@ -511,3 +499,89 @@ func (user *user) applyMessageDeleted(ctx context.Context, update *imap.MessageD
...
+
+	if err := user.db.Write(ctx, func(ctx context.Context, tx *ent.Tx) error {
+		var exists bool
+
+		internalMessageID, err := db.GetMessageIDFromRemoteID(ctx, tx.Client(), update.Message.ID)
+		if err != nil && !ent.IsNotFound(err) {
+			return err
+		} else {
+			exists = true
+		}
```
jameshoulahan, line 514: `if err != nil && !ent.IsNotFound(err)` --> exists will be marked as true if `err == nil || ent.IsNotFound(err)`, so you will attempt to delete nonexistent messages, right? I guess this is a mistake.

LBeernaertProton: Good catch. fixed.

jameshoulahan (approved):

PR #218, opened 2022-11-04
Fix UpdateMessageEvent and rfc822.SetHeaderValue

LBeernaertProton:


rfc822/header.go
```diff
@@ -342,14 +342,18 @@ func SetHeaderValue(literal []byte, key, val string) ([]byte, error) {
 	key = textproto.CanonicalMIMEHeaderKey(key)
 	data := joinLine([]byte(key), []byte(val))
 
+	result := make([]byte, 0, len(data)+len(literal))
 	if !foundFirstEntry {
-		return append(rawHeader, append(data, body...)...), nil
+		result = append(result, rawHeader...)
+		result = append(result, data...)
+		result = append(result, body...)
```
jameshoulahan, line 349: ahhh good catch

seeing as append caused problems in the past... maybe you wanna make it even safer, like this?

```suggestion
		result = bytes.Join([][]byte{rawHeader, data, body}, nil)
```

LBeernaertProton: `result` is a new value, it's practically the same as that statement.

LBeernaertProton: I have also updated the tests to check for this.

jameshoulahan: alright, i'm just paranoid about appends now, but nevermind

internal/backend/connector_updates.go
```diff
@@ -503,78 +504,104 @@ func (user *user) applyMessageDeleted(ctx context.Context, update *imap.MessageD
...
 
-		// delete the message and remove from the mailboxes.
-		if exists {
-			mailboxes, err := db.GetMessageMailboxIDs(ctx, tx.Client(), internalMessageID)
-			if err != nil {
-				return err
+		// compare and see if the literal has changed.
+		onDiskLiteral, err := user.store.Get(internalMessageID)
+		if err != nil {
+			return fmt.Errorf("failed to retrieve literal from cache")
```
jameshoulahan, line 522:
```suggestion
			return fmt.Errorf("failed to retrieve literal from cache: %w", err)
```

might be useful to show the underlying error

LBeernaertProton: Done

internal/backend/connector_updates.go
```diff
@@ -503,78 +504,104 @@ func (user *user) applyMessageDeleted(ctx context.Context, update *imap.MessageD
...
+		if err != nil {
+			return fmt.Errorf("failed to retrieve literal from cache: %w", err)
+		}
+
+		updateLiteral := update.Literal
+		if id, err := rfc822.GetHeaderValue(updateLiteral, ids.InternalIDKey); err == nil {
+			if len(id) == 0 {
+				newLiteral, err := rfc822.SetHeaderValue(updateLiteral, ids.InternalIDKey, internalMessageID.String())
+				if err != nil {
+					log.Debug("failed to set header key, using update literal")
```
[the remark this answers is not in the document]
LBeernaertProton: Done

jameshoulahan (approved):

PR #221, opened 2022-11-10
feat: Return mailbox counts when user added

jameshoulahan:


internal/backend/backend.go
```diff
@@ -122,6 +124,31 @@ func (b *Backend) RemoveUser(ctx context.Context, userID string, removeFiles boo
 	return nil
 }
 
+func (b *Backend) GetCounts(ctx context.Context, userID string) (map[imap.MailboxID]int, error) {
```
LBeernaertProton, line 127: Can we get a bit more descriptive name here?

internal/backend/backend.go
```diff
@@ -122,6 +124,31 @@ func (b *Backend) RemoveUser(ctx context.Context, userID string, removeFiles boo
...
+	b.usersLock.Lock()
+	defer b.usersLock.Unlock()
+
+	return db.ReadResult(ctx, b.users[userID].db, func(ctx context.Context, c *ent.Client) (map[imap.MailboxID]int, error) {
+		counts := make(map[imap.MailboxID]int)
+
+		mailboxes, err := c.Mailbox.Query().All(ctx)
```
LBeernaertProton, line 134: Add select field with ID only to avoid loading more data than necessary.

internal/backend/backend.go
```diff
@@ -122,6 +124,31 @@ func (b *Backend) RemoveUser(ctx context.Context, userID string, removeFiles boo
...
+		if err != nil {
+			return nil, err
+		}
+
+		for _, mailbox := range mailboxes {
+			messages, err := mailbox.QueryUIDs().All(ctx)
```
LBeernaertProton, line 140: Use `.Count()` to avoid loading all the data in memory.

LBeernaertProton (changes requested):

internal/backend/backend.go
jameshoulahan: Good point -- "Counts" means a specific thing on the proton API, but here in gluon it's unclear. Changed to `GetMailboxMessageCounts`.

internal/backend/backend.go
jameshoulahan: done

internal/backend/backend.go
jameshoulahan: done

LBeernaertProton (approved):

PR #228, opened 2022-11-21
fix: Properly handle store flags (when replacing)

jameshoulahan:


internal/state/actions.go
```diff
@@ -436,48 +436,31 @@ func (state *State) actionSetMessageFlags(ctx context.Context, tx *ent.Tx, messa
 	})
 
 	// If setting messages as seen, only set those messages that aren't currently seen.
-	var messagesToApply []imap.MessageID
+	setSeen := make(map[bool][]imap.MessageID)
```
LBeernaertProton, line 439: `make(map[..][..], 2)`

internal/state/actions.go
```diff
@@ -436,48 +436,31 @@ func (state *State) actionSetMessageFlags(ctx context.Context, tx *ent.Tx, messa
...
+	for seen, messageIDs := range setSeen {
+		if err := state.user.GetRemote().SetMessagesSeen(ctx, messageIDs, seen); err != nil {
 			return err
 		}
 	}
 
-	// If setting messages as flagged, only set those messages that aren't currently flagged.
-	if setFlags.ContainsUnchecked(imap.FlagFlaggedLowerCase) {
-		var messagesToApply []imap.MessageID
+	setFlagged := make(map[bool][]imap.MessageID)
```
LBeernaertProton, line 453: make(map[..][..], 2)

LBeernaertProton (changes requested):

internal/state/actions.go
jameshoulahan: even better, replaced with 
```go
	setSeen := map[bool][]imap.MessageID{true: {}, false: {}}
```

internal/state/actions.go
jameshoulahan: replaced with map literal

LBeernaertProton (approved):

PR #230, opened 2022-11-24
feat(GODT-1989): Recovery Mailbox

LBeernaertProton:
Add a Gluon managed mailbox which stores messages which failed to be appended. This is required to handle clients such as Outlook which do not check Append failures when moving message and can result in data loss.

When an Append fails, we insert it into a special mailbox called "RecoveredMessages". Messages in this mailbox support:
* Being copied out
* Being moved out
* Flag modifications
* Being deleted

Any other operation is not allowed.

When messages are moved/copied out of this mailbox they are first created as new messages on the Connector before being added to the target mailbox.

### DIFF
--- a/internal/backend/connector_updates.go
+++ b/internal/backend/connector_updates.go
@@ -67,6 +67,10 @@ func (user *user) apply(ctx context.Context, update imap.Update) error {
 
 // applyMailboxCreated applies a MailboxCreated update.
 func (user *user) applyMailboxCreated(ctx context.Context, update *imap.MailboxCreated) error {
+	if update.Mailbox.ID == ids.GluonInternalRecoveryMailboxRemoteID {
+		return fmt.Errorf("attempting to create protected mailbox (recovery)")
+	}
+
 	if exists, err := db.ReadResult(ctx, user.db, func(ctx context.Context, client *ent.Client) (bool, error) {
 		return db.MailboxExistsWithRemoteID(ctx, client, update.Mailbox.ID)
 	}); err != nil {
@@ -95,6 +99,10 @@ func (user *user) applyMailboxCreated(ctx context.Context, update *imap.MailboxC
 
 // applyMailboxDeleted applies a MailboxDeleted update.
 func (user *user) applyMailboxDeleted(ctx context.Context, update *imap.MailboxDeleted) error {
+	if update.MailboxID == ids.GluonInternalRecoveryMailboxRemoteID {
+		return fmt.Errorf("attempting to delete protected mailbox (recovery)")
+	}
+
 	internalMailboxID, err := db.ReadResult(ctx, user.db, func(ctx context.Context, client *ent.Client) (imap.InternalMailboxID, error) {
 		return db.GetMailboxIDFromRemoteID(ctx, client, update.MailboxID)
 	})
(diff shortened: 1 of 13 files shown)


internal/ids/ids.go
```diff
@@ -73,3 +73,7 @@ func SplitMailboxIDPairSlice(s []MailboxIDPair) ([]imap.InternalMailboxID, []ima
 
 	return internalMailboxIDs, mailboxIDs
 }
+
+const GluonRecoveryMailboxName = "RecoveredMessages"
```
jameshoulahan, line 77: Why no space? `All Mail` has a space so can we have one for this too?

internal/backend/connector_updates.go
```diff
@@ -279,6 +300,10 @@ func (user *user) applyMessagesCreated(ctx context.Context, update *imap.Message
 
 // applyMessageMailboxesUpdated applies a MessageMailboxesUpdated update.
 func (user *user) applyMessageMailboxesUpdated(ctx context.Context, update *imap.MessageMailboxesUpdated) error {
+	if slices.Contains(update.MailboxIDs, ids.GluonInternalRecoveryMailboxRemoteID) {
+		return fmt.Errorf("attempting to move messages into proceted mailbox (recovery)")
```
jameshoulahan, line 304:
```suggestion
		return fmt.Errorf("attempting to move messages into protected mailbox (recovery)")
```

internal/backend/connector_updates.go
```diff
@@ -179,6 +195,11 @@ func (user *user) applyMessagesCreated(ctx context.Context, update *imap.Message
 
 	if err := user.db.Read(ctx, func(ctx context.Context, client *ent.Client) error {
 		for _, message := range update.Messages {
+			if slices.Contains(message.MailboxIDs, ids.GluonInternalRecoveryMailboxRemoteID) {
+				logrus.Errorf("attempting to import messages into proceted mailbox (recovery), skipping")
```
jameshoulahan, line 199:
```suggestion
				logrus.Errorf("attempting to import messages into protected mailbox (recovery), skipping")
```

internal/backend/user.go
```diff
@@ -52,6 +55,24 @@ func newUser(
 		return nil, err
 	}
 
+	// Create recovery mailbox if it does not exist
+	recoveryMBox, err := db.WriteResult(ctx, database, func(ctx context.Context, tx *ent.Tx) (*ent.Mailbox, error) {
+		mboxFlags := imap.NewFlagSet(imap.FlagSeen, imap.FlagFlagged, imap.FlagDeleted)
```
jameshoulahan, line 60: do we need to support \Seen and \Flagged? Why not just \Deleted?

LBeernaertProton: We should be able to change the flagged status of the message. I see no reason why we can't support this.

internal/ids/ids.go
LBeernaertProton: Fixed.

internal/backend/connector_updates.go
LBeernaertProton: Fixed.

internal/backend/connector_updates.go
LBeernaertProton: Fixed.

internal/backend/user.go
LBeernaertProton: If these flags not supported, does this mean we should not allow this to be modified then?

jameshoulahan: good question, i don't even know what gluon does if a client tries to interact with flags we don't list as supported 🤔 i guess in theory we support any flag, even those we don't announce as supported.

LBeernaertProton: I don't think we have anything that's checking  for that. TBD in a separate ticket/MR

jameshoulahan (approved):

PR #234, opened 2022-11-29
fix: Drop non-UTF-8 commands

jameshoulahan:


cuthix (approved):

LBeernaertProton (commented):
I'm concerned that this might break for search commands which have other encodings.

jameshoulahan:
> I'm concerned that this might break for search commands which have other encodings.

Indeed, you are right, but it turns out this was already broken! I reproduced in a test (commented out, with TODO) and opened a ticket for it.

cuthix (approved):

PR #236, opened 2022-11-30
fix: RemoveMessagesFromMailbox should only remove from remote if need be

jameshoulahan:
When adding a message to a mailbox, it can happen that the message is already there. In that case, we remove it and re-add it. However, that change only needs to be done locally; it's useless to do it on the remote. Further, it can even be dangerous to do on the remote, as the remote might handle this as a permanent deletion.

### DIFF
--- a/internal/state/mailbox.go
+++ b/internal/state/mailbox.go
@@ -147,8 +147,14 @@ func (m *Mailbox) GetMessagesWithoutFlagCount(flag string) int {
 
 func (m *Mailbox) AppendRegular(ctx context.Context, literal []byte, flags imap.FlagSet, date time.Time) (imap.UID, error) {
 	var appendIntoDrafts bool
+
+	attr, err := m.Attributes(ctx)
+	if err != nil {
+		return 0, err
+	}
+
 	// Force create message when appending to drafts so that IMAP clients can create new draft messages.
-	if !strings.EqualFold(m.name, "Drafts") {
+	if !attr.Contains(imap.AttrDrafts) {
 		internalIDString, err := rfc822.GetHeaderValue(literal, ids.InternalIDKey)
 		if err != nil {
 			return 0, err
--- a/internal/state/mailbox.go
+++ b/internal/state/mailbox.go
@@ -19,7 +19,11 @@ import (
 )
 
 type Mailbox struct {
-	mbox *ent.Mailbox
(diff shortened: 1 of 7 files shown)


cuthix:
I think this is realistic scenario and we should fix it as well: https://github.com/ProtonMail/gluon/pull/237

LBeernaertProton:
If we do things locally without consulting the connector, we are risking to do illegal operations. E.g: It's not allowed to append or add messages to All Mail. If we don't call the connector. We can't validate this.

internal/state/mailbox.go
```diff
@@ -153,53 +153,65 @@ func (m *Mailbox) GetMessagesWithoutFlagCount(flag string) int {
...
-	internalIDString, err := rfc822.GetHeaderValue(literal, ids.InternalIDKey)
-	if err != nil {
-		return 0, err
-	}
-
-	if len(internalIDString) > 0 {
-		msgID, err := imap.InternalMessageIDFromString(internalIDString)
+	var appendIntoDrafts bool
+	// Force create message when appending to drafts so that IMAP clients can create new draft messages.
+	if !strings.EqualFold(m.mbox.Name, "Drafts") {
```
cuthix, line 158: can we base it on special use flag `\Drafts`, so connector can decide what name it's gonna be and controls the behaviour. I am not sure if we allowed localisation in Bridge yet.

tests/draft_test.go
```diff
@@ -39,3 +42,73 @@ func TestDraftScenario(t *testing.T) {
...
+				regexp.QuoteMeta("* 1 FETCH (BODY[] {82}\r\nX-Pm-Gluon-Id: ") +
+					".*" +
+					regexp.QuoteMeta("\r\nTo: 3@3.pm)\r\n\r\nThis is body\r\n)"),
+			)
+
+			c.C("A004 FETCH 1 (BODY.PEEK[])")
+			raw := c.read()
+			require.Regexp(t, fetchUpdated, string(raw))
+			c.OK("A004")
+		}
```
cuthix, line 112: Does `Append + Expunge` work as well?

LBeernaertProton: Yes. All imports are new messages.  I will add a test case.

LBeernaertProton: Done

cuthix (approved):

PR #242, opened 2022-12-06
feat(GODT-1970): Soft limits

LBeernaertProton:
Add limit checks for mailbox count, messages per mailbox, UID and UIDValidity so that we don't run into issues related exceeding the maximum representable size for each of these data types (uint32).

### DIFF
--- a/limits/imap.go
+++ b/limits/imap.go
@@ -1,0 +1,82 @@
+package limits
+
+import (
+	"errors"
+	"fmt"
+	"github.com/ProtonMail/gluon/imap"
+	"math"
+)
+
+// IMAP contains configurable upper limits that can be enforced by the Gluon server.
+type IMAP struct {
+	maxMailboxCount           int
+	maxMessageCountPerMailbox int
+	maxUIDValidity            int
+	maxUID                    int
+}
+
+func (i IMAP) CheckMailBoxCount(mailboxCount int) error {
+	if mailboxCount >= i.maxMailboxCount {
+		return ErrMaxMailboxCountReached
+	}
+
+	return nil
+}
+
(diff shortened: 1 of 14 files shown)


limits/imap.go
```diff
@@ -0,0 +1,82 @@
...
+
+	if nextUIDCount > i.maxUID || nextUIDCount < int(existingUID) {
+		return ErrMaxUIDReached
+	}
+
+	return nil
+}
+
+func (i IMAP) CheckUIDValidity(uid imap.UID) error {
+	if int(uid) == i.maxUIDValidity {
```
cuthix, line 47: and `>=` here ?

limits/imap.go
```diff
@@ -0,0 +1,82 @@
...
+// IMAP contains configurable upper limits that can be enforced by the Gluon server.
+type IMAP struct {
+	maxMailboxCount           int
+	maxMessageCountPerMailbox int
+	maxUIDValidity            int
+	maxUID                    int
+}
+
+func (i IMAP) CheckMailBoxCount(mailboxCount int) error {
+	if mailboxCount == i.maxMailboxCount {
```
cuthix, line 19: why not using `>=` here?

LBeernaertProton: mailboxes are always create 1 at a time

limits/imap.go
LBeernaertProton: mailboxes are always create 1 at a time

LBeernaertProton: But just in case, changed.

limits/imap.go
LBeernaertProton: But just in case, changed.

cuthix (approved):

PR #244, opened 2022-12-12
fix(GODT-1896): Do not remove subscribed status for deleted mailboxes

LBeernaertProton:
Ensure that we do not delete subscriptions entries for deleted mailboxes without an explicit unsubscribe event (RFC 3501 Section 6.3.9).

If a new mailbox is created with the same name as previously deleted and subscribed mailbox, the subscription data is overwritten to point to the newly created mailbox.


internal/state/match.go
```diff
@@ -3,11 +3,11 @@ package state
 import (
 	"context"
 	"fmt"
+	"github.com/ProtonMail/gluon/internal/db"
```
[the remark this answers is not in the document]
LBeernaertProton: Done

internal/db/subscription.go
```diff
@@ -0,0 +1,67 @@
...
+import (
+	"context"
+	"github.com/ProtonMail/gluon/imap"
+	"github.com/ProtonMail/gluon/internal/db/ent"
+	"github.com/ProtonMail/gluon/internal/db/ent/subscription"
+	"github.com/ProtonMail/gluon/internal/ids"
+)
+
+func AddSubscription(ctx context.Context, tx *ent.Tx, mboxName string, mboxID ids.MailboxIDPair) error {
+	count, err := tx.Subscription.Update().Where(subscription.NameEqualFold(mboxName)).SetMailboxID(mboxID.InternalID).SetRemoteID(mboxID.RemoteID).Save(ctx)
```
jameshoulahan, line 12: I'm happy with how you did this, no need to change it, but just fyi we might be able to use upsert here (https://entgo.io/docs/feature-flags#upsert)

LBeernaertProton: Noted!

LBeernaertProton:
> 

Good catch.  Fixed and unit test added.

internal/state/match.go
```diff
@@ -18,16 +18,23 @@ type Match struct {
 	Atts      imap.FlagSet
 }
 
+type MatchMailbox struct {
```
jameshoulahan, line 21: Could you make this private (`matchMailbox`) so as to not confuse with the other public "pseudo-mailbox" (`AppendOnlyMailbox`)?

LBeernaertProton: Done

jameshoulahan (approved):

PR #245, opened 2022-12-13
fix(GODT-2207): Fix missing utf7 encoding for responses

LBeernaertProton:
Some responses which contain mailbox names were not encoded into utf7. This can lead to duplicate subscription folders and IMAP clients being unable to list the contents of a mailbox.

### DIFF
--- a/internal/session/handle_status.go
+++ b/internal/session/handle_status.go
@@ -49,7 +49,7 @@ func (s *Session) handleStatus(ctx context.Context, tag string, cmd *proto.Statu
 			}
 		}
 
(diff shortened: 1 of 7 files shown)


internal/session/handle_status.go
```diff
@@ -49,7 +49,11 @@ func (s *Session) handleStatus(ctx context.Context, tag string, cmd *proto.Statu
 			}
 		}
 
-		ch <- response.Status().WithMailbox(nameUTF8).WithItems(items...)
+		if err != nil {
```
jameshoulahan, line 52: maybe forgot to remove?

LBeernaertProton: Done

PR #260, opened 2023-01-17
feat(GODT-2252): Recover from missing cache files

LBeernaertProton:
Allow commands such as fetch and search to re-download literals from the connector if they are no longer available.

To ensure parallel fetches which download the same literal don't cause file system issue with writing multiple files, we wrap the store implementation with a layer that controls access to the message with a RWLock.


internal/state/mailbox_fetch.go
```diff
@@ -111,12 +112,26 @@ func (m *Mailbox) Fetch(ctx context.Context, seq *proto.SequenceSet, attributes
...
+				if err := m.state.user.GetStore().Set(msg.ID.InternalID, connectorLiteral); err != nil {
+					logrus.Errorf("Failed to store download message from connector: %v", err)
+					return firstErr
+				}
+
+				logrus.Debugf("Message %v downloaded and stored ", msg.ID.InternalID.ShortID())
+				literal = connectorLiteral
+			} else {
+				literal = l
+			}
```
jameshoulahan, line 134: Just a thought, could `state.getLiteral` transparently fetch/save the message if necessary, thereby hiding the complexity from the mailbox? This block of code is duplicated in mailbox_fetch.go and mailbox_search.go when it could probably be done once in `state.getLiteral` instead.

LBeernaertProton: Good point, fixed.

internal/state/state.go
```diff
@@ -621,8 +621,32 @@ func (state *State) endIdle() {
...
+	var literal []byte
+
+	storeLiteral, firstErr := state.user.GetStore().Get(messageID.InternalID)
+	if firstErr != nil {
+		logrus.Debugf("Failed load %v from store, attempting to download from connector", messageID.InternalID.ShortID())
+
+		connectorLiteral, err := state.user.GetRemote().GetMessageLiteral(ctx, messageID.RemoteID)
+		if err != nil {
+			logrus.Errorf("Failed to download message from connector: %v", err)
+			return nil, firstErr
```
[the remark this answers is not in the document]
LBeernaertProton: Network retry should be handled in the connector.

jameshoulahan (dismissed):

jameshoulahan (approved):

PR #267, opened 2023-01-30
GODT-2224- Changes required to reduce  ram usage [BREAKING CHANGE]

LBeernaertProton:


rfc822/header.go
```diff
@@ -311,6 +311,25 @@ func (h *Header) applyOffset(start *headerEntry, offset int) {
...
+	if _, err := b.ReadFrom(reader); err != nil {
+		return nil, err
+	}
+
+	return b.Bytes(), nil
+}
+
+// SetHeaderValueNoCopy is a helper method that sets a header value in a message literal.
+// It does not check whether the existing value already exists.
+func SetHeaderValueNoCopy(literal []byte, key, val string) (io.Reader, int, error) {
```
jameshoulahan, line 332: Maybe add info to the docstring about what "NoCopy" actually means (which values aren't being copied? What are the behaviour differences between this version and the above version?)

store/disk.go
```diff
@@ -46,64 +49,168 @@ func NewOnDiskStore(path string, pass []byte, opt ...Option) (Store, error) {
...
+
+				return
+			}
+
+			// Decrypt read bytes.
+			decryptBuffer = decryptBuffer[:0] // Reset slice.
+
+			decrypted, err := c.gcm.Open(decryptBuffer, nonce, readBuffer[0:bytesRead], nil)
+			if err != nil {
+				writer.CloseWithError(fmt.Errorf("failed to decypt block (offset:%v): %w", totalBytesRead, err))
```
jameshoulahan, line 185:
```suggestion
				writer.CloseWithError(fmt.Errorf("failed to decrypt block (offset: %v): %w", totalBytesRead, err))
```

store/write_controlled_store.go
```diff
@@ -18,13 +19,18 @@ type WriteControlledStore struct {
...
-	lockPool   []*syncRef
+	lockPool   sync.Pool
 }
 
 func NewWriteControlledStore(impl Store) *WriteControlledStore {
 	return &WriteControlledStore{
 		impl:       impl,
 		entryTable: make(map[imap.InternalMessageID]*syncRef),
+		lockPool: sync.Pool{
+			New: func() interface{} {
```
jameshoulahan, line 30:
```suggestion
			New: func() any {
```

with go 1.18 `interface{}` is pretty much deprecated, use `any` wherever you can

store/write_controlled_store.go
```diff
@@ -34,19 +40,15 @@ func (w *WriteControlledStore) acquireSyncRef(id imap.InternalMessageID) *syncRe
...
-
-		if len(w.lockPool) != 0 {
-			s = w.lockPool[0]
-			s.counter = 1
-			w.lockPool = w.lockPool[1:]
-		} else {
-			s = &syncRef{counter: 1}
+		v, ok := w.lockPool.Get().(*syncRef)
+		if !ok {
+			panic("Invalid Type Cast")
```
jameshoulahan, line 45:
```suggestion
			panic("invalid type cast")
```

little style nitpick, go panics should all be lowercase (for some reason)

store/disk.go
```diff
@@ -46,64 +49,168 @@ func NewOnDiskStore(path string, pass []byte, opt ...Option) (Store, error) {
...
 	if c.sem != nil {
 		c.sem.Lock()
 		defer c.sem.Unlock()
 	}
 
-	enc, err := os.ReadFile(filepath.Join(c.path, messageID.String()))
-	if err != nil {
-		return nil, err
+	nonce := make([]byte, c.gcm.NonceSize())
+	if _, err := rand.Read(nonce); err != nil {
```
jameshoulahan, line 65: I've been using `crypto.RandomToken` in the past to have a "standardized" way of generating random tokens in bridge. I think under the hood they do the same thing as you, but to be safe, it might be good to stick to their implementation in case they improve it in future. Thoughts?

LBeernaertProton: I would address this in a follow MR to avoid scope creep :)

store/store.go
```diff
@@ -1,10 +1,13 @@
 package store
 
-import "github.com/ProtonMail/gluon/imap"
+import (
+	"github.com/ProtonMail/gluon/imap"
+	"io"
+)
 
 type Store interface {
 	Get(messageID imap.InternalMessageID) ([]byte, error)
```
jameshoulahan, line 9: Quick question, for symmetry, might we want to have streaming interface for both Get and Set? Might help with reducing copies down the line? Or not worth it you think?

tests/draft_test.go
```diff
@@ -53,6 +53,7 @@ func TestDraftSavedAgain(t *testing.T) {
 		c.C("A002 SELECT Drafts").OK("A002")
 
 		_ = s.messageCreated("user", mailboxID, []byte("To: 3@3.pm"), time.Now())
+		s.flush("user")
```
jameshoulahan, line 56: Not relevant to this PR but i wonder why/how you realized you had to add this? Did your changes affect the runtime behaviour in such a way as to require a flush? Or was it already flaky and you just hit a failure here at some point?

store/store.go
LBeernaertProton: It currently doesn't have any benefits as we always need to decompress the whole thing in memory. There is a bit more work required to make sure that everything works correctly with the  RWLock system that sits over it.

I agree that it woud look better consistent, but it's not a high priority atm.

tests/draft_test.go
LBeernaertProton: This was a test that would occasionally fail in the past.

store/write_controlled_store.go
LBeernaertProton: Done

store/write_controlled_store.go
LBeernaertProton: Done

store/disk.go
LBeernaertProton: Done

store/disk.go
```diff
@@ -46,64 +49,168 @@ func NewOnDiskStore(path string, pass []byte, opt ...Option) (Store, error) {
...
 
-		b = dec
+		// Encrypt the compressed block.
+		encryptedBlock = encryptedBlock[:0] // Reset slice.
+		encrypted := c.gcm.Seal(encryptedBlock, nonce, compressedBlock[0:bytesRead], nil)
+		encryptedLen := bytesRead + encryptionOverhead
+
+		// Write to disk.
+		if bytesWritten, err := file.Write(encrypted[0:encryptedLen]); err != nil || bytesWritten != encryptedLen {
+			return fmt.Errorf("failed to write blcock to disk: %w", err)
```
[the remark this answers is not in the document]
LBeernaertProton: done

rfc822/header.go
LBeernaertProton: Done

jameshoulahan (approved):

PR #275, opened 2023-02-13
feat(GODT-2201) Golang IMAP parser POC

LBeernaertProton:


imap/command/search.go
```diff
@@ -0,0 +1,383 @@
...
+	}
+
+	return fmt.Sprintf("SEARCH CHARSET=%v %v", charsetStr, xslices.Map(s.Keys, func(v SearchKey) string {
+		return v.SanitizedString()
+	}))
+}
+
+type SearchCommandParser struct {
+	charset        string
+	charsetAllowed bool
```
jameshoulahan, line 43: Instead of having this internal parser state, why not just check if the first keyword is CHARSET, if so parse it, and then continue parsing keys?

imap/command/search.go
```diff
@@ -0,0 +1,383 @@
...
+	scp.charsetAllowed = true
+
+	var keys []SearchKey
+
+	for {
+		if !p.Check(parser.TokenTypeSP) {
+			break
+		}
+
+		key, err := scp.parseSearchKey(p)
```
jameshoulahan, line 58: if you have parsed a charset, you should look up the decoder from IANA package and use it to decode any search terms, shouldn't you?

LBeernaertProton: I would rather deal with this while handling the search command. The important thing is that we apply the decoder before we use the search term.

imap/command/search.go
LBeernaertProton: Done

imap/command/search.go
LBeernaertProton: Additionally, it would be good to keep the original search parameters after parsing for debugging.

jameshoulahan (approved):

PR #277, opened 2023-02-14
feat(GODT-2201): IMAP Empty Value Commands

LBeernaertProton:
Implements the following IMAP commands which have no attached values:

* Capability
* Check
* Close
* Done
* Expunge
* Idle
* Logout
* Noop
* Unselect

### DIFF
--- a/imap/command/done_test.go
+++ b/imap/command/done_test.go
@@ -1,0 +1,31 @@
+package command
+
+import (
+	"bytes"
+	"github.com/ProtonMail/gluon/imap/parser"
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+func TestParser_DoneCommand(t *testing.T) {
+	input := toIMAPLine(`DONE`)
+	s := parser.NewScanner(bytes.NewReader(input))
+	p := NewParser(s)
+
+	expected := Command{Tag: "", Payload: &DoneCommand{}}
+
+	cmd, err := p.Parse()
+	require.NoError(t, err)
+	require.Equal(t, expected, cmd)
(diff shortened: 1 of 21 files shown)


imap/command/done_test.go
```diff
@@ -0,0 +1,22 @@
...
+
+import (
+	"bytes"
+	"github.com/ProtonMail/gluon/imap/parser"
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+func TestParser_DoneCommand(t *testing.T) {
+	input := toIMAPLine(`tag DONE`)
```
jameshoulahan, line 11: done is sent without a tag

imap/command/logout.go
```diff
@@ -0,0 +1,22 @@
...
+
+import (
+	"fmt"
+	"github.com/ProtonMail/gluon/imap/parser"
+)
+
+type LogoutCommand struct{}
+
+func (l LogoutCommand) String() string {
+	return fmt.Sprintf("NOOP")
```
jameshoulahan, line 11: logout

cuthix: how could it pass the test ?

imap/command/done_test.go
LBeernaertProton: Good catch. Done (no pun intented)

imap/command/logout.go
LBeernaertProton: This is just for debug logging, not for the actual parsing.

Fixed.

jameshoulahan (approved):

PR #290, opened 2023-02-16
Replace C++ IMAP Parser with new  Go Version

LBeernaertProton:


internal/session/command.go
```diff
@@ -37,43 +31,46 @@ func (s *Session) startCommandReader(ctx context.Context, del string) <-chan com
...
-			}
+			s.inputCollector.Reset()
 
-			s.logIncoming(string(line), xslices.Map(maps.Keys(literals), func(k string) string {
-				return fmt.Sprintf("%v: '%s'", k, literals[k])
-			})...)
+			cmd, err := parser.Parse()
+			s.logIncoming(string(s.inputCollector.Bytes()))
+			if err != nil {
+				parserError, ok := err.(*rfcparser.Error) //nolint:errorlint
```
jameshoulahan, line 42: I'd prefer you use errors.As here rather, any reason not to?

internal/session/command.go
```diff
@@ -37,43 +31,46 @@ func (s *Session) startCommandReader(ctx context.Context, del string) <-chan com
...
-			tag, cmd, err := parse(line, literals, del)
-			if err != nil {
 				reporter.MessageWithContext(ctx,
 					"Failed to parse IMAP command",
-					reporter.Context{"error": err},
+					reporter.Context{"error": err, "cmd": parser.LastParsedCommand()},
 				)
-			} else if cmd.GetStartTLS() != nil {
+			} else {
+				logrus.Debug(cmd.SaniztedString())
```
jameshoulahan, line 67: Typo

internal/session/handle.go
```diff
@@ -52,47 +52,46 @@ func (s *Session) handleOther(
...
-		cmd.GetIdSet() != nil,
-		cmd.GetNoop() != nil:
+		*command.Capability,
+		*command.IDGet,
+		*command.IDSet,
+		*command.Noop:
 		return s.handleAnyCommand(ctx, tag, cmd, ch)
 
 	case
-		cmd.GetAuth() != nil,
```
jameshoulahan, line 67: What's the current behaviour if an IMAP server _does_ send us the auth command? Will we return NO or BAD? Could we implement some minimal parsing to ensure we can detect AUTHENTICATE and return NO?

internal/session/handle_append.go
```diff
@@ -3,32 +3,30 @@ package session
...
-	"github.com/emersion/go-imap/utf7"
 )
 
-func (s *Session) handleAppend(ctx context.Context, tag string, cmd *proto.Append, ch chan response.Response) error {
+func (s *Session) handleAppend(ctx context.Context, tag string, cmd *command.Append, ch chan response.Response) error {
 	profiling.Start(ctx, profiling.CmdTypeAppend)
 	defer profiling.Stop(ctx, profiling.CmdTypeAppend)
 
-	nameUTF8, err := utf7.Encoding.NewDecoder().String(cmd.GetMailbox())
+	nameUTF8, err := s.decodeMailboxName(cmd.Mailbox)
```
jameshoulahan, line 18: Any reason not to auto-decode mailbox names inside the parser? We have a jira ticket for that, it might be nice to do now.

internal/session/handle_expunge.go
```diff
@@ -29,7 +29,7 @@ func (s *Session) handleExpunge(ctx context.Context, tag string, cmd *proto.Expu
 	return response.Ok(tag).WithMessage("EXPUNGE"), nil
 }
 
-func (s *Session) handleUIDExpunge(ctx context.Context, tag string, cmd *proto.UIDExpunge, mailbox *state.Mailbox, ch chan response.Response) (response.Response, error) {
+func (s *Session) handleUIDExpunge(ctx context.Context, tag string, cmd *command.UIDExpungeCommand, mailbox *state.Mailbox, ch chan response.Response) (response.Response, error) {
```
jameshoulahan, line 32: Didn't we want to remove the Command suffix on all command types? Maybe you forgot this one

internal/session/handle_search.go
```diff
@@ -3,16 +3,16 @@ package session
...
-	"github.com/ProtonMail/gluon/internal/parser/proto"
 	"github.com/ProtonMail/gluon/internal/response"
 	"github.com/ProtonMail/gluon/internal/state"
 	"github.com/ProtonMail/gluon/profiling"
 	"golang.org/x/text/encoding"
 	"golang.org/x/text/encoding/ianaindex"
 )
 
-func (s *Session) handleSearch(ctx context.Context, tag string, cmd *proto.Search, mailbox *state.Mailbox, ch chan response.Response) (response.Response, error) {
+func (s *Session) handleSearch(ctx context.Context, tag string, cmd *command.Search, mailbox *state.Mailbox, ch chan response.Response) (response.Response, error) {
```
jameshoulahan, line 15: Haven't finished reviewing yet, but note to self: now that we should be able to support parsing arbitrary bytes, let's ensure we have a test for SEARCH with non-utf-8 stuff (which should have failed before but should pass now)

internal/session/handle_store.go
```diff
@@ -21,16 +21,16 @@ func (s *Session) handleStore(ctx context.Context, tag string, cmd *proto.Store,
...
 	}
 
-	flags, err := validateStoreFlags(cmd.GetFlags())
+	flags, err := validateStoreFlags(cmd.Flags)
 	if err != nil {
 		return response.Bad(tag).WithError(err), nil
 	}
 
-	if err := mailbox.Store(ctx, cmd.GetSequenceSet(), cmd.GetAction().GetOperation(), flags); errors.Is(err, state.ErrNoSuchMessage) {
+	if err := mailbox.Store(ctx, cmd, flags); errors.Is(err, state.ErrNoSuchMessage) {
```
jameshoulahan, line 33: Why leak the command.Store type through to the backend? I liked it more how it was, where Store would just operate on SeqSet and Operation

internal/session/handle.go
LBeernaertProton: Right now it's a parser error, will return the error to the client with bad

internal/state/mailbox.go
```diff
@@ -301,7 +301,7 @@ func (m *Mailbox) Copy(ctx context.Context, seq *proto.SequenceSet, name string)
 // Move moves the messages represented by the given sequence set into the mailbox with the given name.
 // If the context is a UID context, the sequence set refers to message UIDs.
 // If no items are moved the response object will be nil.
-func (m *Mailbox) Move(ctx context.Context, seq *proto.SequenceSet, name string) (response.Item, error) {
+func (m *Mailbox) Move(ctx context.Context, seq []command.SeqRange, name string) (response.Item, error) {
```
jameshoulahan, line 304: Wait why does Move take a SeqRange and not a SeqSet? 🤔

LBeernaertProton: `SeqSet == []SeqRange`

internal/session/handle_store.go
LBeernaertProton: Done

internal/session/handle_expunge.go
LBeernaertProton: Done

internal/session/command.go
LBeernaertProton: Done

internal/session/command.go
LBeernaertProton: Done

internal/state/mailbox.go
jameshoulahan: Ah okay, must have misread, i thought when i left that comment that Copy accepted SeqSet and Move accepted SeqRange, and wondered why they didn't accept the same type. My bad.

internal/session/handle_append.go
jameshoulahan: Could you store the raw/decoded alongside each other then? Or could you add a method on cmd.Mailbox like `cmd.Mailbox.Decode()`? I really want to try avoid all this decoding code-duplication.

internal/session/handle.go
jameshoulahan: How much work would it be to enable us to return NO here? BAD usually indicates protocol-level errors whereas NO indicates feature-level errors. I'd prefer if we could implement the minimal amount of parsing necessary to be able to reject AUTHENTICATE requests with NO rather than BAD.

LBeernaertProton: Have to parse the full Authenticate command to get this recognized properly.

internal/session/handle_search.go
LBeernaertProton: There is GODT-2165 which covers this topic. We can work on that after this MR.

internal/session/handle_append.go
LBeernaertProton: I would really like to avoid having to pass the mailbox separator character down the parsing pipeline as it is not really required to complete parsing.

Your suggestion for `cmd.Mailbox.Decode` is no different than what we have at the moment. `decodeMailboxName` does a bit more than just decoding from utf7.

internal/session/handle_search.go
jameshoulahan: Thing is, we had a test that would assert that non-utf-8 search would throw an error and that was deleted -- couldn't you just for this MR assert that no error is thrown? Then when we do GODT-2165 we can extend the test to make sure decoding works properly? Just wanting to make sure we don't get any surprises with this MR.

LBeernaertProton: This error no longer exists. It doesn't make sense anymore.

LBeernaertProton: Renabled disabled test covered by GODT-2165 to verify that decoding of non-ascii chars works.

jameshoulahan (approved):

PR #291, opened 2023-02-20
fix: Unsubscribed mailboxes should always be listed with LIST

jameshoulahan:


LBeernaertProton:
Looks good, but could you add a test that captures the original issue?

PR #295, opened 2023-02-21
fix: No error if failing to remove deleted messages from store

jameshoulahan:


internal/backend/user.go
```diff
@@ -101,7 +101,11 @@ func newUser(
 	}
 
 	if err := user.deleteAllMessagesMarkedDeleted(ctx); err != nil {
-		return nil, err
+		logrus.WithError(err).Error("Failed to remove deleted messages")
```
LBeernaertProton, line 104: I think it might be better to make sure the database is emptied of the messages even if the store deletion fails.

Right now if the store deletion fails, the deleted message wills till remain in the database.

jameshoulahan: I don't think that's the case -- we do the DB delete first inside deleteAllMessagesMarkedDeleted

LBeernaertProton (approved):

PR #296, opened 2023-02-21
feat: Silently create missing messages on update

jameshoulahan:


internal/backend/connector_updates.go
```diff
@@ -598,16 +598,21 @@ func (user *user) applyMessageDeleted(ctx context.Context, update *imap.MessageD
...
-	if err := user.db.Write(ctx, func(ctx context.Context, tx *ent.Tx) error {
-		internalMessageID, err := db.GetMessageIDFromRemoteID(ctx, tx.Client(), update.Message.ID)
-		if err != nil {
-			if ent.IsNotFound(err) {
-				log.Debugf("Message %v no longer exists, skipping", update.Message.ID.ShortID())
-				return nil
-			}
-			return err
-		}
+	internalMessageID, err := db.ReadResult(ctx, user.db, func(ctx context.Context, client *ent.Client) (imap.InternalMessageID, error) {
```
LBeernaertProton, line 601: While this solves the issue. I think it would be better to have this as a separate event that replicates this.

This way we can reduce the surface of accidental errors.

jameshoulahan: I've added a bool to control whether missing messages should optionally be created

LBeernaertProton (approved):

LBeernaertProton (approved):

PR #297, opened 2023-02-21
feat: Expose methods to query ErrNoSuchMessage / ErrNoSuchMailbox

jameshoulahan:


errors.go
```diff
@@ -0,0 +1,18 @@
...
+package gluon
+
+import (
+	"errors"
+
+	"github.com/ProtonMail/gluon/internal/state"
+)
+
+// IsNoSuchMessage returns true if the error is ErrNoSuchMessage.
+func IsNoSuchMessage(err error) bool {
```
LBeernaertProton, line 11: Can you verify this with at test. I have a suspicion that this will not work as we are reporting ent errors to the connector.

jameshoulahan: done, it works as expected

LBeernaertProton (approved):

PR #298, opened 2023-02-21
feat(GODT-2201): RFC5322 Rewrite DateTime parser in Go

LBeernaertProton:


internal/parser/rfc5322_date_time_test.go
```diff
@@ -1,6 +1,7 @@
 package parser
 
 import (
+	"github.com/ProtonMail/gluon/rfc5322"
```
jameshoulahan, line 4: goimports pls

jameshoulahan (dismissed):

internal/parser/rfc5322_date_time_test.go
LBeernaertProton: Done

jameshoulahan (approved):

PR #299, opened 2023-02-22
GODT-2385: Store fallback mechanism 

LBeernaertProton:
* Add header to store files
* Allow fallback reader to read old files.

### DIFF
--- a/store/disk.go
+++ b/store/disk.go
@@ -5,6 +5,7 @@ import (
 	"crypto/aes"
 	"crypto/cipher"
 	"crypto/rand"
+	"encoding/binary"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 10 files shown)


store/disk.go
```diff
@@ -279,6 +325,38 @@ func (*OnDiskStoreBuilder) Delete(path, userID string) error {
...
+
+func uint32ToByteArray(i uint32) []byte {
+	var b []byte
+	sh := (*reflect.SliceHeader)(unsafe.Pointer(&b))
+	sh.Len = 4
+	sh.Cap = 4
+	sh.Data = uintptr(unsafe.Pointer(&i))
+
+	return b[:]
+}
```
jameshoulahan, line 350: probably better to use `encoding/binary` here (https://pkg.go.dev/encoding/binary)

LBeernaertProton: I found that, but that is variant based decoding.  I need a fixed value so I can always read the same amount of bytes in one go.

LBeernaertProton: Sorry. I misread the documentation. Fixed.

jameshoulahan (dismissed):

jameshoulahan (approved):

PR #305, opened 2023-02-27
fix(GODT-2399): Defer updated message deletion

LBeernaertProton:
Ensure that when a message gets update, we defer deletion of the message so that it remains accessible until all snapshots that have this message have been released.

### DIFF
--- a/internal/backend/connector_updates.go
+++ b/internal/backend/connector_updates.go
@@ -677,7 +677,9 @@ func (user *user) applyMessageUpdated(ctx context.Context, update *imap.MessageU
 					stateUpdates = append(stateUpdates, updates...)
 				}
 
-				if err := db.DeleteMessages(ctx, tx, internalMessageID); err != nil {
+				// We need change the old remote id as it will break our table constraint otherwise and everything
+				// will silently fail.
+				if err := db.MarkMessageAsDeletedAndAssignRandomRemoteID(ctx, tx, internalMessageID); err != nil {
 					return err
 				}
 			}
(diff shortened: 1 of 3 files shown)


internal/backend/connector_updates.go
```diff
@@ -699,8 +701,10 @@ func (user *user) applyMessageUpdated(ctx context.Context, update *imap.MessageU
 					InternalID:  newInternalID,
 				}
 
-				if _, err := db.CreateMessages(ctx, tx, request); err != nil {
+				if m, err := db.CreateMessages(ctx, tx, request); err != nil {
 					return err
+				} else if len(m) == 0 {
+					return fmt.Errorf("no messages where inserted")
```
jameshoulahan, line 707:
```suggestion
					return fmt.Errorf("no messages were inserted")
```

internal/db/message.go
```diff
@@ -560,6 +560,15 @@ func MarkMessageAsDeleted(ctx context.Context, tx *ent.Tx, messageID imap.Intern
 	return nil
 }
 
+func MarkMessageAsDeletedAndAssignRandomRemoteID(ctx context.Context, tx *ent.Tx, messageID imap.InternalMessageID) error {
+	randomID := imap.MessageID(fmt.Sprintf("DELETED-%v", imap.NewInternalMessageID().String()))
```
jameshoulahan, line 564:
```suggestion
	randomID := imap.MessageID(fmt.Sprintf("DELETED-%v", imap.NewInternalMessageID()))
```

should auto-call String() since it satisfies Stringer interface

internal/backend/connector_updates.go
LBeernaertProton: Done

internal/db/message.go
LBeernaertProton: Done

jameshoulahan (approved):

PR #306, opened 2023-02-27
fix(GODT-2400): Allow state updates to be applied if command fails

LBeernaertProton:
In the case there is an unexpected issue with the IMAP command, still allow pending state updates to be applied.

### DIFF
--- a/internal/session/handle.go
+++ b/internal/session/handle.go
@@ -221,11 +221,18 @@ func (s *Session) handleSelectedCommand(
 
 	return s.state.Selected(ctx, func(mailbox *state.Mailbox) error {
 		okResponse, err := s.handleWithMailbox(ctx, tag, cmd, mailbox, ch)
-		if err != nil {
-			return err
+
+		// Allow state updates to be applied if the command failed. It might resolve
+		// some invalid state problem.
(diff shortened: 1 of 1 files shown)


internal/session/handle.go
```diff
@@ -221,11 +221,19 @@ func (s *Session) handleSelectedCommand(
...
+		// Allow state updates to be applied if the command failed. It might resolve
+		// some invalid state problem.
+		flushErr := flush(ctx, mailbox, false, ch)
+		if flushErr != nil {
+			if err != nil {
+				return fmt.Errorf("%w (flush err:%v)", err, flushErr)
+			}
+
+			return flushErr
 		}
```
jameshoulahan, line 234:
```suggestion
		// Allow state updates to be applied if the command failed. It might resolve
		// some invalid state problem.
		if flushErr := flush(ctx, mailbox, false, ch); flushErr != nil {
			if err != nil {
				return fmt.Errorf("%w (flush err: %v)", err, flushErr)
			}

			return flushErr
		}
```

since we have multiple errors here, it would be nice to scope the flush error just to this block so it doesn't get confusing

LBeernaertProton: Done

PR #314, opened 2023-03-03
feat: Support answered flag in message updates

jameshoulahan:


imap/update_message_labels_updated.go
```diff
@@ -11,29 +11,34 @@ type MessageMailboxesUpdated struct {
...
+	return fmt.Sprintf("seen=%v, flagged=%v, draft=%v, answered=%v", f.Seen, f.Flagged, f.Draft, f.Answered)
 }
 
-func NewMessageMailboxesUpdated(messageID MessageID, mailboxIDs []MailboxID, seen, flagged, draft bool) *MessageMailboxesUpdated {
+func NewMessageMailboxesUpdated(messageID MessageID, mailboxIDs []MailboxID, flags MessageCustomFlags) *MessageMailboxesUpdated {
 	return &MessageMailboxesUpdated{
 		updateWaiter: newUpdateWaiter(),
 		MessageID:    messageID,
 		MailboxIDs:   mailboxIDs,
-		Seen:         seen,
```
LBeernaertProton, line 24: While these changes are correct. I think it would be better if we normalized the flag manipulation interface so that we can send any type of flag back and forth.

LBeernaertProton (changes requested):

PR #322, opened 2023-03-08
fix(GODT-2442): Use rename instead of remove for db files

LBeernaertProton:
Unfortunately, closing the Ent database clients doesn't guarantee that all the SQLite db files are closed. This causes issues on Windows where the files are locked, making it impossible to remove and re-add the same user.

### DIFF
--- a/internal/db/db.go
+++ b/internal/db/db.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"github.com/ProtonMail/gluon/internal/utils"
+	"github.com/google/uuid"
 	"io/fs"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


internal/db/db.go
```diff
@@ -132,8 +137,42 @@ func NewDB(dir, userID string) (*DB, bool, error) {
...
 func DeleteDB(dir, userID string) error {
-	return os.Remove(getDatabasePath(dir, userID))
+	// Rather than deleting the files immediately move them to a directory to be cleaned up later.
+	deferredDeletePath := getDeferredDeleteDBPath(dir)
+
+	if err := os.MkdirAll(deferredDeletePath, 0o700); err != nil {
+		return fmt.Errorf("failed to create deferred delete dir: %w", err)
+	}
+
+	matchingFiles, err := filepath.Glob(fmt.Sprintf("%v/%v*", dir, userID))
```
cuthix, line 151: would this be better?

```suggestion
	matchingFiles, err := filepath.Glob(filepath.Join(dir, userID))
```

LBeernaertProton: Good catch. Fixed.

cuthix (approved):

PR #332, opened 2023-03-24
feat(GODT-2500): Add panic handlers everywhere.

cuthix:


wait/wg.go
```diff
@@ -2,14 +2,31 @@ package wait
...
+	HandlePanic()
+}
+
 type Group struct {
-	wg sync.WaitGroup
+	wg           sync.WaitGroup
+	panicHandler PanicHandler
+}
+
+func (wg *Group) SetPanicHandler(panicHandler PanicHandler) {
```
LBeernaertProton, line 14: Can we have this as part of the constructor so it's not possible to change this after it is created?

queue/queued_channel.go
```diff
@@ -18,7 +18,15 @@ type QueuedChannel[T any] struct {
 	closed atomicBool // Should use atomic.Bool once we use Go 1.19!
 }
 
-func NewQueuedChannel[T any](chanBufferSize, queueCapacity int) *QueuedChannel[T] {
+type PanicHandler interface {
```
LBeernaertProton, line 21: Can we have this in a different package/folder. I feel it would be better to have this self contained rather than being part of the queue.

xmichelo (approved):

PR #339, opened 2023-04-24
fix(GODT-2582): Improve message hashing for recovery dedup

LBeernaertProton:
Improve the message hashing by ensuring we hash the body of each part of the message rather than just the entire blob. This will ensure we can handle cases where the only thing that changes between messages is the part identifiers.

### DIFF
--- a/internal/utils/message_hashmap.go
+++ b/internal/utils/message_hashmap.go
@@ -1,9 +1,8 @@
 package utils
 
 import (
-	"crypto/sha256"
-	"encoding/hex"
 	"github.com/ProtonMail/gluon/imap"
+	"github.com/ProtonMail/gluon/rfc822"
 	"sync"
 )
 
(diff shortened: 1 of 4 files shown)


internal/utils/message_hashmap.go
```diff
@@ -24,15 +23,11 @@ func NewMessageHashesMap() *MessageHashesMap {
 // Insert inserts the hash of the current message literal into the map and return true if an existing value was already
 // present.
 func (m *MessageHashesMap) Insert(id imap.InternalMessageID, literal []byte) (bool, error) {
-	hash := sha256.New()
-
-	if _, err := hash.Write(literal); err != nil {
+	literalHashStr, err := rfc822.GetMessageHash(literal)
+	if err != nil {
```
cuthix, line 27: if `err!=nil` then try `literalHash` as before?

LBeernaertProton: We need to parse here again sadly, since the result of the previous parsing is lost.

cuthix (approved):

PR #349, opened 2023-05-11
fix(GODT-2623): log IMAP login failure as error.

rlejeune74:


internal/backend/backend.go
```diff
@@ -248,6 +248,7 @@ func (b *Backend) getUserID(ctx context.Context, username string, password []byt
 			defer b.loginWG.Done()
 			atomic.StoreInt32(&b.loginErrorCount, 0)
 		})
+		logrus.WithError(ErrLoginBlocked).Error("failed IMAP login")
```
LBeernaertProton, line 251: Isn't `ErrLoginBlocked` the wrong error message here?

rlejeune74: That's matching the one returned the line after ;)

rlejeune74: > 	ErrLoginBlocked = errors.New("too many login attempts")

LBeernaertProton: This will only log when we reached failed login attempts limit. It won't log a login failure.

rlejeune74: oh yeah true....

rlejeune74: Done

internal/backend/backend.go
```diff
@@ -250,6 +250,8 @@ func (b *Backend) getUserID(ctx context.Context, username string, password []byt
 		})
 
 		return "", ErrLoginBlocked
+	} else {
+		logrus.Errorf("Incorrect login credentials. Please try again, Attempt %d", count)
```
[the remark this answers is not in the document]
rlejeune74: That matches the old log the CS relies on.
+ after 3 failed attempt you get blocked so this could lead to better understanding once the user get blocked

LBeernaertProton (approved):

PR #359, opened 2023-05-26
Remove Ent Part 1

LBeernaertProton:
Hide ent implementation behind an interface


db/errors.go
```diff
@@ -0,0 +1,14 @@
+package db
+
+import "errors"
+
+var ErrNotFound = errors.New("value not found")
+var ErrTransactionFailed = errors.New("value not found")
```
rlejeune74, line 6: should be something more like "transaction failed"

db/ops_message.go
```diff
@@ -0,0 +1,94 @@
...
+
+	MessageExistsWithRemoteID(ctx context.Context, id imap.MessageID) (bool, error)
+
+	GetMessage(ctx context.Context, id imap.InternalMessageID) (*Message, error)
+
+	GetTotalMessageCount(ctx context.Context) (int, error)
+
+	GetMessageRemoteID(ctx context.Context, id imap.InternalMessageID) (imap.MessageID, error)
+
+	//GetMessageWithRemoteID(ctx context.Context, id imap.MessageID) (*Message, error)
```
rlejeune74, line 22: dead code ?

db/errors.go
LBeernaertProton: Good catch. Fixed.

db/ops_message.go
LBeernaertProton: Yep. Fixed.

rlejeune74 (approved):

PR #361, opened 2023-06-02
chore: Increase test timeout.

rlejeune74:


LBeernaertProton:
Can you provide some tests that would benefit from the timeout increase?

PR #370, opened 2023-06-27
feat(GODT-2522): migration recovery

LBeernaertProton:
When migration fails auto-recreate the database.

### DIFF
--- a/internal/db_impl/db_impl.go
+++ b/internal/db_impl/db_impl.go
@@ -1,10 +1,15 @@
 package db_impl
 
 import (
+	"context"
 	"github.com/ProtonMail/gluon/db"
 	"github.com/ProtonMail/gluon/internal/db_impl/sqlite3"
 )
 
 func NewSQLiteDB(options ...sqlite3.Option) db.ClientInterface {
 	return sqlite3.NewBuilder(options...)
+}
+
+func TestUpdateDBVersion(ctx context.Context, dbPath, userID string, version int) error {
+	return sqlite3.TestUpdateDBVersion(ctx, dbPath, userID, version)
 }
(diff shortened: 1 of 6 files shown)


internal/db_impl/sqlite3/client.go
```diff
@@ -289,3 +289,22 @@ func pathExists(path string) (bool, error) {
 func getDatabaseConn(dir, userID, path string) string {
 	return fmt.Sprintf("file:%v?cache=shared&_fk=1&_journal=WAL", path)
 }
+
+func TestUpdateDBVersion(ctx context.Context, dbPath, userID string, version int) error {
```
cuthix, line 293: should this be in `client_test.go` ?

internal/db_impl/db_impl.go
```diff
@@ -1,10 +1,15 @@
...
+	"context"
 	"github.com/ProtonMail/gluon/db"
 	"github.com/ProtonMail/gluon/internal/db_impl/sqlite3"
 )
 
 func NewSQLiteDB(options ...sqlite3.Option) db.ClientInterface {
 	return sqlite3.NewBuilder(options...)
 }
+
+func TestUpdateDBVersion(ctx context.Context, dbPath, userID string, version int) error {
```
cuthix, line 13: should this be in `db_impl_test.go` ?

internal/db_impl/sqlite3/client.go
LBeernaertProton: The logic that verifies if the db is corrupted and erases it in `backend.go`. To check for this I have to start the server add the user and observer the behavior. I need access to a protected method inside the SQLite implementation.

Alternatively, I need to add a method to the `db` interface to write the database version, which I don't think is a good idea 👀

internal/db_impl/db_impl.go
LBeernaertProton: See comment above

cuthix (approved):

PR #383, opened 2023-07-10
limits/imap.go: Prevent integer overflow on 32 bit platforms

### LINKED ISSUE #382: 32 bit support
The file limits/imap.go uses `MaxUint32` which is too big for the `int` type on 32 bits and causes builds on those architectures to fail. This seems to be a known issue based on proton-bridge's BUILDS.md file, but I don't see any effort made to solve it. Is this specific number required, or can something else 32-bit-friendly be used?

benthetechguy:
Fixes #382
Since the (signed) int type is used, MaxUInt32 is too big for use on 32 bit platforms. Changing to MaxInt32 solves this problem.

### DIFF
--- a/limits/imap.go
+++ b/limits/imap.go
@@ -4,20 +4,21 @@ import (
 	"errors"
 	"fmt"
 	"math"
+	"math/bits"
 
 	"github.com/ProtonMail/gluon/imap"
 )
 
 // IMAP contains configurable upper limits that can be enforced by the Gluon server.
 type IMAP struct {
-	maxMailboxCount           int
-	maxMessageCountPerMailbox int
-	maxUIDValidity            int
-	maxUID                    int
+	maxMailboxCount           int64
+	maxMessageCountPerMailbox int64
+	maxUIDValidity            int64
+	maxUID                    int64
 }
 
 func (i IMAP) CheckMailBoxCount(mailboxCount int) error {
-	if mailboxCount >= i.maxMailboxCount {
+	if int64(mailboxCount) >= i.maxMailboxCount {
 		return ErrMaxMailboxCountReached
 	}
 
@@ -25,9 +26,9 @@ func (i IMAP) CheckMailBoxCount(mailboxCount int) error {
 }
 
 func (i IMAP) CheckMailBoxMessageCount(existingCount int, newCount int) error {
-	nextMessageCount := existingCount + newCount
+	nextMessageCount := int64(existingCount) + int64(newCount)
 
-	if nextMessageCount > i.maxMessageCountPerMailbox || nextMessageCount < existingCount {
+	if nextMessageCount > i.maxMessageCountPerMailbox || nextMessageCount < int64(existingCount) {
 		return ErrMaxMailboxMessageCountReached
 	}
 
@@ -35,9 +36,9 @@ func (i IMAP) CheckMailBoxMessageCount(existingCount int, newCount int) error {
 }
 
 func (i IMAP) CheckUIDCount(existingUID imap.UID, newCount int) error {
-	nextUIDCount := int(existingUID) + newCount
+	nextUIDCount := int64(existingUID) + int64(newCount)
 
-	if nextUIDCount > i.maxUID || nextUIDCount < int(existingUID) {
+	if nextUIDCount > i.maxUID || nextUIDCount < int64(existingUID) {
 		return ErrMaxUIDReached
 	}
 
@@ -45,7 +46,7 @@ func (i IMAP) CheckUIDCount(existingUID imap.UID, newCount int) error {
 }
 
 func (i IMAP) CheckUIDValidity(uid imap.UID) error {
-	if int(uid) >= i.maxUIDValidity {
+	if int64(uid) >= i.maxUIDValidity {
 		return ErrMaxUIDValidityReached
 	}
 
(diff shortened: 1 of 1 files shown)


limits/imap.go
```diff
@@ -54,10 +54,10 @@ func (i IMAP) CheckUIDValidity(uid imap.UID) error {
 
 func DefaultLimits() IMAP {
 	return IMAP{
-		maxMailboxCount:           math.MaxUint32,
-		maxMessageCountPerMailbox: math.MaxUint32,
-		maxUIDValidity:            math.MaxUint32,
-		maxUID:                    math.MaxUint32,
+		maxMailboxCount:           math.MaxInt32,
```
LBeernaertProton: The value still needs to be `MaxUint32 (4294967295)`. What I suggested originally is to replace the data type of these variables from`int` to `uint32`. This should work for 32bit systems.

benthetechguy: Ah, misread your suggestion; fixed.

limits/imap.go
```diff
@@ -27,7 +27,7 @@ func (i IMAP) CheckMailBoxCount(mailboxCount int) error {
 func (i IMAP) CheckMailBoxMessageCount(existingCount int, newCount int) error {
 	nextMessageCount := existingCount + newCount
 
-	if nextMessageCount > i.maxMessageCountPerMailbox || nextMessageCount < existingCount {
+	if uint32(nextMessageCount) > i.maxMessageCountPerMailbox || nextMessageCount < existingCount {
```
LBeernaertProton: Hmm. This brings up an interesting problem....

Since every call to `len()` in go is now going to return a 32bit singed integer (`int`) these checks will always fail and risk running into overflow territory.

We need to do 2 things here.

1)  The calculation of next message count and UID needs to be done in 64 bits.
```
nextMessageCount := int64(existingCount) + int64(count)
```

2) The comparisons also need to be promoted to int64  e.g.:
```
nextMessageCount > int64(i.maxMessageCountPerMailbox) || nextMessageCount < int64(existingCount)
```

3) Finally we need to set different limits for 32bit and 64bit compilation.
On 32bit systems the limits need to be set to `MaxInt32` since we have lot of other calculation that rely on the output of `len()` which only returns int32 values.

For 64bit, we can leave the `MaxUint32` since we can represent the calculations in the full range.

benthetechguy:
Is there any reason we can't just use MaxInt32 for limits on all systems, like my original changes did?

LBeernaertProton:
> Is there any reason we can't just use MaxInt32 for limits on all systems, like my original changes did?

The maximum range of values we can represent with `MaxInt32` is `[0, 2147483647]`

The maximum range of values we can represent with `MaxUint32` is `[0, 4294967295]`

The current setup works, since `int` on 64bit system is equal to `int64`, which can safely express all the range of operations we need to perform to make the calculations.

While your original change compiles, it has some problems:

* 64bit systems are now handicapped (max range of values reduced by 2) 
* Once we hit `MaxInt32` values, the checks will always fail on 32bit system without due to signed integer overflow

benthetechguy:
So you're okay with 32 bit platforms having half the max range as long as 64 bit ones don't also?

limits/imap.go
benthetechguy: If we're upgrading all of the `int`s to `int64`s, why do we have to use `MaxInt32` on 32 bit platforms? What's preventing the use of `MaxUint32` on 32 bit platforms if we're using 64 bit integers for all the calculations?

benthetechguy: P.S. forgive my ignorance throughout this PR discussion as I have no golang experience whatsoever outside of building and packaging existing code

LBeernaertProton:
> So you're okay with 32 bit platforms having half the max range as long as 64 bit ones don't also?

Yes, 64bit should have full range. I see this as compromise for now. 

There is much more work required to ensure that one can represent the max range of uint32 correctly across the entire code base in 32bit mode. But at least it would not block people from building under 32bits.

limits/imap.go
LBeernaertProton: @benthetechguy if we can verify that`int` type on 32bits can represent a 64bit integer value, no changes need to be applied. However, based on past experiences, `int` types on 32bit system as limited to a 32bit integer only. On 64bit the `int` type is  `64bits` long.

You can check the size of the type with this snippet:
```
package main

import (
	"fmt"
	"unsafe"
)

func main() {
	fmt.Println(unsafe.Sizeof(int(0))) 
}
```

benthetechguy: I believe `int` on 32 bit platforms is indeed limited to the `int32` range, but you can use `int64` explicitly to get the full 64 bit range by (I think) the compiler splitting it across two instructions. Since we're making all these values `int64`s, we shouldn't have to worry about overflows, right?

LBeernaertProton: The problem is this built-in function `len() int` and other similar functions. The current snapshot state code relies on that. 

Even though we can make the database queries return the correct types, we would need to review the whole code base to ensure that that we are not losing precision range when counting things.

benthetechguy: If we make `MaxInt32` the limit on 32 bit platforms, wouldn't it be within the range of `int32`, preventing overflow?

LBeernaertProton: Yes.  That is correct. See my comment: https://github.com/ProtonMail/gluon/pull/383#issuecomment-1633635549

benthetechguy: Why do we need to represent all of these variables as `int64` if we're never going to overflow using the limit of `MaxInt32` on 32 bit platforms?

LBeernaertProton: Due to signed integer overflow: `MaxInt32 +1 = -2147483648`. On 64bit system it's enough to promote the types to int , since it is in itself already 64bits. 

On 32bits we need to do the calculations in 64bit space so that `MaxInt32+1` does not overflow.

benthetechguy: Okay, thank you for helping me understand.

limits/imap.go
```diff
@@ -10,63 +10,71 @@ import (
...
 		return ErrMaxUIDValidityReached
 	}
 
 	return nil
 }
 
 func DefaultLimits() IMAP {
+	var maxInt int64
+	// Use MaxInt32 instead of MaxUint32 on 32 bit platforms
+	if uint64(^uintptr(0)) == ^uint64(0) {
```
LBeernaertProton: I'm not sure this will compile on 32bits. I just tried `if uint64(^uintptr(0)) == ^uint32(0) ` on my machine that failed to compile.

benthetechguy:
I got it from StackOverflow; it's supposed to return true on 64 bits and false on 32. It compiled fine on my 32 bit test machine, haven't tried 64.

limits/imap.go
benthetechguy: It compiled fine on my 32 bit test VM. Anyway, changed to using `UintSize` from `math/bits`.

LBeernaertProton:
Looks good now. Could you amend the commit to `fix: Prevent IMAP Limits integer overflow on 32 bit platforms`

benthetechguy:
I don't know if you were notified, but I did it three days ago.

LBeernaertProton (approved):

PR #393, opened 2023-08-16
fix(GODT-2873): Add `Wait` to `QueuedChannel`

LBeernaertProton:
Required so that we can wait until the queued channel go-routine finishes during the Gluon Watcher closing.

### DIFF
--- a/watcher/watcher.go
+++ b/watcher/watcher.go
@@ -44,4 +44,5 @@ func (w *Watcher[T]) Send(event T) bool {
 
 func (w *Watcher[T]) Close() {
 	w.eventCh.CloseAndDiscardQueued()
+	w.eventCh.Wait()
 }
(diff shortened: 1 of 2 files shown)


watcher/watcher.go
```diff
@@ -44,4 +44,5 @@ func (w *Watcher[T]) Send(event T) bool {
 
 func (w *Watcher[T]) Close() {
 	w.eventCh.CloseAndDiscardQueued()
+	w.eventCh.Wait()
```
cuthix, line 47: is it safe to add `Wait` directly to `Close`? shouldn't we use different function `CloseAndWait` ?

LBeernaertProton: Yes this is safe and preferred. That way we can opt in where required without changing the existing code & behavior.

cuthix (approved):

PR #400, opened 2023-11-14
Feat/godt 2576

LBeernaertProton:


tests/store_test.go
```diff
@@ -2,6 +2,9 @@ package tests
 
 import (
 	"fmt"
+	"github.com/ProtonMail/gluon/imap"
+	"golang.org/x/exp/slices"
+	"strings"
```
rlejeune74, line 6: File is not goimported

LBeernaertProton: Done

rlejeune74 (approved):

PR #401, opened 2023-12-06
fix(GODT-3153): Do not take into account full address when hasing mes…

rlejeune74:
…sage.

### DIFF
--- a/rfc822/hash.go
+++ b/rfc822/hash.go
@@ -8,6 +8,7 @@ import (
 	"mime/quotedprintable"
 	"strings"
 
+	"github.com/ProtonMail/gluon/rfc5322"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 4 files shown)


rfc822/hash.go
```diff
@@ -152,3 +153,15 @@ func hashBody(writer io.Writer, body []byte, mimeType MIMEType, encoding string)
 
 	return err
 }
+
+func getAddresses(fieldAddr string) string {
+	addrList, err := rfc5322.ParseAddressList(fieldAddr)
+	if err != nil {
+		return ""
```
LBeernaertProton: Should we maybe just return the full string here?

LBeernaertProton (changes requested):

LBeernaertProton (approved):

PR #405, opened 2024-02-26
feat(GODT-3199): improve logging.

cuthix:
* add package log field
* add IDs to when fetching body

### DIFF
--- a/internal/state/mailbox_fetch.go
+++ b/internal/state/mailbox_fetch.go
@@ -33,6 +33,7 @@ func (m *Mailbox) Fetch(ctx context.Context, cmd *command.Fetch, ch chan respons
 		needsLiteral bool
 		wantUID      bool
 		setSeen      bool
+		isBodyFetch  bool
 	)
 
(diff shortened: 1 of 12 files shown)


internal/state/mailbox_fetch.go
```diff
@@ -78,6 +81,7 @@ func (m *Mailbox) Fetch(ctx context.Context, cmd *command.Fetch, ch chan respons
 			operations = append(operations, fetchBodyStructure)
 		case *command.FetchAttributeBodySection:
 			needsLiteral = true
+			isBodyFetch = true
```
cuthix, line 84: I think it would be better if we loop over items and set `isBodyFetch=true` only if there are no items or at least one item is non-header :thinking:

xmichelo (dismissed):

xmichelo (approved):

internal/state/mailbox_fetch.go
cuthix: checked main email clients, there might be double counts but logs will be silent eventually.

PR #406, opened 2024-03-01
feat(BRIDGE-23):go 1.21 updates, govulncheck github action and package update.

xmichelo:


CuriouslyCurious:
It should be noted that since this PR updates `go-sqlite3` to a [version greater than `v1.14.19`](https://github.com/mattn/go-sqlite3/issues/1164#issuecomment-1975022901) this project now compiles on musl-based system like Alpine Linux just fine, thus leading to the potentiality that [`proton-bridge` could compile on musl-based systems](https://github.com/ProtonMail/proton-bridge/issues/333). 

I don't have any familiarity with go beyond running `go build` so if more needs to be done than just bumping versions here and upstream in `proton-bridge` I am unsure.

cuthix (approved):

PR #407, opened 2024-04-22
fix(BRIDGE-10): escaping db path when creating connection

ElectroNafta:
Switched to using a parsed URL encoded path for DB connection, in order to avoid sqlite issues with special characters in the db path.

### DIFF
--- a/internal/db_impl/sqlite3/client_test.go
+++ b/internal/db_impl/sqlite3/client_test.go
@@ -1,0 +1,47 @@
+package sqlite3
+
+import (
+	"database/sql"
+	"fmt"
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+func TestClient_DBConnectionSpecialCharacterPath(t *testing.T) {
+	dbDirs := []string{
+		"#test",
+		"test_test",
+		"test#test#test",
+		"test$test$test",
+	}
+
+	testingDir := t.TempDir()
+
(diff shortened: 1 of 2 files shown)


internal/db_impl/sqlite3/client_test.go
```diff
@@ -0,0 +1,57 @@
...
+	"os/user"
+	"path/filepath"
+	"testing"
+)
+
+const (
+	TestingDirectoryRoot = "/tmp/test_db_sqlite"
+)
+
+func TestClient_db_connection_special_characters_path(t *testing.T) {
```
xmichelo: We don't use snake case. `TestClientDBConnectionSpecialCharacterPath` or `TestClient_DBConnectionSpecialCharacterPath`

internal/db_impl/sqlite3/client_test.go
```diff
@@ -0,0 +1,57 @@
...
+		"test$test$test",
+	}
+
+	curUser, err := user.Current()
+	if err != nil {
+		fmt.Println("Could not get current current user, error: ", err)
+		t.FailNow()
+	}
+
+	testingDir := filepath.Join(curUser.HomeDir, TestingDirectoryRoot)
```
xmichelo: Don't create file, even temporary, at the root of the user's home. use `os.TempPath()`.

xmichelo (approved):

PR #414, opened 2024-10-02
feat(BRIDGE-218): observability support & metrics; reviewed remaining…

ElectroNafta:
… sentry events

### DIFF
--- a/observability/metrics/metrics.go
+++ b/observability/metrics/metrics.go
@@ -1,0 +1,56 @@
+package metrics
+
+import "time"
+
+const schemaName = "bridge_gluon_errors_total"
+const schemaVersion = 1
+
+func generateGluonErrorMetric(errorType string) map[string]interface{} {
+	return map[string]interface{}{
+		"Name":      schemaName,
+		"Version":   schemaVersion,
+		"Timestamp": time.Now().Unix(),
+		"Data": map[string]interface{}{
+			"Value": 1,
+			"Labels": map[string]string{
+				"errorType": errorType,
+			},
+		},
+	}
+}
+
+func GenerateFailedParseIMAPCommandMetric() map[string]interface{} {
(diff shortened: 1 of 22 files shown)


option.go
```diff
@@ -243,3 +244,16 @@ func (w withDBClient) config(builder *serverBuilder) {
...
+
+type withObservabilitySender struct {
+	sender observability.Sender
+}
+
+func (w withObservabilitySender) config(builder *serverBuilder) {
+	builder.observabilitySender = w.sender
+}
+
+func WithObsSender(sender observability.Sender, imapErrorType, messageErrorType, otherErrorType int) Option {
```
xmichelo: 1. Please rename to `WithObservabilitySender()`
2. You're not calling this yet. Waiting for some backend changes?

observability/metrics/metrics.go
```diff
@@ -0,0 +1,73 @@
...
+
+func GenerateAppendToDraftsMustNotReturnExistingRemoteID() map[string]interface{} {
+	return generateGluonErrorMetric("appendToDraftsReturnedExistingRemoteID")
+}
+
+func GenerateDatabaseMigrationFailed() map[string]interface{} {
+	return generateGluonErrorMetric("databaseMigrationFailed")
+}
+
+func GenerateAllMetrics() []map[string]interface{} {
```
xmichelo: What's the purpose of this unused function?

ElectroNafta: It's just for testing, which we'll do via Bridge.

I suppose it's better to put it there instead of Gluon. I'll change this

option.go
ElectroNafta: 1. ✅ 
2. We wont be calling this in Gluon. It will be used by Bridge so it can pass the observability service down to Gluon.

xmichelo (approved):

PR #418, opened 2024-11-20
chore: update golangci-lint to v1.62.0 and fix reported issues.

xmichelo:


internal/session/handle_examine.go
```diff
@@ -36,8 +36,8 @@ func (s *Session) handleExamine(ctx context.Context, tag string, cmd *command.Ex
 		}
 
 		ch <- response.Flags().WithFlags(flags)
-		ch <- response.Exists().WithCount(imap.SeqID(mailbox.Count()))
-		ch <- response.Recent().WithCount(uint32(mailbox.GetMessagesWithFlagCount(imap.FlagRecent)))
+		ch <- response.Exists().WithCount(imap.SeqID(uint32(mailbox.Count())))                       //nolint:gosec
```
cuthix: instead of turninig off all gosec can we specify only the one issue we want to ignore?

xmichelo: Updated following slack dock. G115 check is now disabled.

ElectroNafta (approved):

cuthix (approved):

PR #423, opened 2025-01-10
feat(BRIDGE-288): handle previously existing message in create events…

ElectroNafta:
Changed the message creation handler logic. If a previously existing message is encountered in a create event, instead of simply appending new mailboxes to the message we now fully update it. 

This includes comparing the message literal, as well the actual difference in assigned mailboxes.

### DIFF
--- a/internal/backend/connector_updates.go
+++ b/internal/backend/connector_updates.go
@@ -204,6 +204,9 @@ func (user *user) applyMessagesCreated(ctx context.Context, update *imap.Message
 	messageForMBox := make(map[imap.InternalMailboxID][]db.MessageIDPair)
 	mboxInternalIDMap := make(map[imap.MailboxID]imap.InternalMailboxID)
 
+	// for existing messages - we fully update them
+	var messagesToUpdate []*imap.MessageCreated
+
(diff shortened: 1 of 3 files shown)


internal/backend/connector_updates.go
```diff
@@ -339,9 +345,23 @@ func (user *user) applyMessagesCreated(ctx context.Context, update *imap.Message
 				}
 			}
 		}
+		return err
 	}
 
-	return err
+	for _, message := range messagesToUpdate {
```
LBeernaertProton, line 351: I don't recall the exact location or i may be misremembering, but shouldn't this be executed in the same location where `messagesToCreate` is handled?

ElectroNafta: It will cause a deadlock. 

Both main DB operations in `applyMessageCreated` and `applyMessageUpdated` are wrapped with  `userDBWrite(ctx, user, func(ctx context.Context, tx db.Transaction) ([]state.Update, error)` which ends up acquiring a shared lock. 

I intentionally put the forward to `applyMessageUpdated` after  the wrapped function.

xmichelo (approved):

PR #446, opened 2025-12-19
fix(BRIDGE-449): Add interval-triggered untagged OK responses for long lasting commands

ElectroNafta:


internal/session/cmdwatcher/watcher.go
```diff
@@ -0,0 +1,120 @@
...
+	startTime        time.Time
+	sanitizedPayload string
+}
+
+type SendProgressMessageFn func(msg string) error
+
+type Service struct {
+	ctx                   context.Context
+	rwLock                sync.RWMutex
+	commandWatch          map[uuid.UUID]commandMetadata
```
xmichelo: Maybe there something I'm missing here, but why do we need to track multiple commands ? We will not be accepting new command from the client until the current one finishes, right?

ElectroNafta: You're correct on that front, thanks for the comment 🙏

xmichelo (dismissed):

xmichelo (approved):

PR #455, opened 2026-03-10
feat(BRIDGE-464): add the ability to limit the amount of IMAP connect…

annatar:
…ions via client

### DIFF
--- a/imap/connectioncounter/connectioncounter.go
+++ b/imap/connectioncounter/connectioncounter.go
@@ -20,7 +20,8 @@ type RollingCounter struct {
 
 	log *logrus.Entry
 
-	newConnectionThreshold int
+	observabilityConnectionThreshold int
+	connectionLimitThreshold         int
 
 	numberOfBuckets int
 	buckets         []int
@@ -35,17 +36,19 @@ type RollingCounter struct {
 	connProvider        openConnectionProvider
 }
 
-func NewRollingCounter(newConnectionTreshold, numberOfBuckets int, bucketRotationInterval time.Duration) *RollingCounter {
+func NewRollingCounter(connectionLimitThreshold, observabilityConnectionThreshold, numberOfBuckets int, bucketRotationInterval time.Duration) *RollingCounter {
 	log := logrus.WithFields(logrus.Fields{
-		"pkg":       "gluon/rollingcounter",
-		"threshold": newConnectionTreshold,
+		"pkg":                              "gluon/rollingcounter",
+		"connectionLimitThreshold":         connectionLimitThreshold,
+		"observabilityConnectionThreshold": observabilityConnectionThreshold,
 	})
 
 	rc := &RollingCounter{
-		newConnectionThreshold: newConnectionTreshold,
-		numberOfBuckets:        numberOfBuckets,
-		bucketRotationInterval: bucketRotationInterval,
-		log:                    log,
+		observabilityConnectionThreshold: observabilityConnectionThreshold,
+		connectionLimitThreshold:         connectionLimitThreshold,
+		numberOfBuckets:                  numberOfBuckets,
+		bucketRotationInterval:           bucketRotationInterval,
+		log:                              log,
 	}
 
 	return rc
@@ -74,20 +77,18 @@ func (rc *RollingCounter) Start(ctx context.Context, obsSender observability.Sen
 }
 
 func (rc *RollingCounter) run() {
-	rc.wg.Add(1)
-	go func() {
-		defer rc.wg.Done()
+	rc.wg.Go(func() {
 		for {
 			select {
 			case <-rc.ctx.Done():
 				return
 
 			case <-rc.bucketRotationTicker.C:
-				rc.thresholdCheck()
+				rc.observabilityThresholdCheck()
 				rc.onBucketRotationTick()
 			}
 		}
-	}()
+	})
 }
 
 func (rc *RollingCounter) Stop() {
@@ -105,9 +106,10 @@ func (rc *RollingCounter) withBucketLock(fn func()) {
 	fn()
 }
 
-func (rc *RollingCounter) thresholdCheck() {
+func (rc *RollingCounter) observabilityThresholdCheck() {
 	rollingCount := rc.GetRollingCount()
-	if rollingCount < rc.newConnectionThreshold {
+
+	if rollingCount < rc.observabilityConnectionThreshold {
 		return
 	}
 
@@ -135,9 +137,24 @@ func (rc *RollingCounter) NewConnection() {
 }
 
 func (rc *RollingCounter) GetRollingCount() int {
+	return rc.getRollingCounterSafe()
+}
+
+func (rc *RollingCounter) OverConnectionLimitThreshold() bool {
 	rc.bucketLock.Lock()
 	defer rc.bucketLock.Unlock()
 
+	return rc.getRollingCountUnsafe() >= rc.connectionLimitThreshold
+}
+
+func (rc *RollingCounter) getRollingCounterSafe() int {
+	rc.bucketLock.Lock()
+	defer rc.bucketLock.Unlock()
+
+	return rc.getRollingCountUnsafe()
+}
+
+func (rc *RollingCounter) getRollingCountUnsafe() int {
 	var rollingCount int
 	for _, count := range rc.buckets {
 		rollingCount += count
--- a/imap/connectioncounter/connectioncounter_test.go
+++ b/imap/connectioncounter/connectioncounter_test.go
@@ -1,7 +1,6 @@
 package connectioncounter_test
 
 import (
-	"context"
 	"sync"
 	"testing"
 	"time"
@@ -24,9 +23,9 @@ type mockObsSender struct {
 	lastOpenConns, lastNewlyOpenedConns int
 }
 
-func (m *mockObsSender) AddMetrics(_ ...map[string]interface{}) {}
+func (m *mockObsSender) AddMetrics(_ ...map[string]any) {}
 
-func (m *mockObsSender) AddDistinctMetrics(_ interface{}, _ ...map[string]interface{}) {}
+func (m *mockObsSender) AddDistinctMetrics(_ any, _ ...map[string]any) {}
 
 func (m *mockObsSender) AddIMAPConnectionsExceededThresholdMetric(openConns, newlyOpenedConns int) {
 	m.mu.Lock()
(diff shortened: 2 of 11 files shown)


imap/connectionlimiter/client.go
```diff
@@ -0,0 +1,30 @@
...
+	ClientAppleMail   Client = "apple-mail"
+	ClientOutlook     Client = "outlook"
+	ClientThunderbird Client = "thunderbird"
+	ClientUnknown     Client = "unknown"
+)
+
+func normalizeClientKey(id imap.IMAPID) Client {
+	name := strings.TrimSpace(strings.ToLower(id.Name))
+	switch {
+	case strings.Contains(name, "mac") || strings.Contains(name, "os x") || strings.Contains(name, "apple"):
```
ElectroNafta: This will map other cases as well, such as `Outlook for Mac`.

Use the same principle as we do on Bridge i.e. 
``` go
case strings.Contains(userAgent, "mac") && strings.Contains(userAgent, "mail"):
```

imap/connectioncounter/connectioncounter.go
```diff
@@ -138,10 +138,21 @@ func (rc *RollingCounter) GetRollingCount() int {
...
 
-	var rollingCount int
+	return rc.getRollingCountUnlocked()
+}
+
+func (rc *RollingCounter) OverThreshold() bool {
+	rc.bucketLock.Lock()
+	defer rc.bucketLock.Unlock()
+
+	return rc.getRollingCountUnlocked() >= rc.newConnectionThreshold
```
ElectroNafta: We shouldn't rely on the `newConnectionThreshold`. We need a separate value dedicated for this. 

The method name should also correspond to something like OverNewConnectionTreshold. 

But this also means that the following `newConnectionThreshold` should be renamed to something like `newConnectionObservabilitySendThreshold` or simillar.

imap/connectioncounter/connectioncounter.go
```diff
@@ -138,10 +138,21 @@ func (rc *RollingCounter) GetRollingCount() int {
 	rc.bucketLock.Lock()
 	defer rc.bucketLock.Unlock()
 
-	var rollingCount int
+	return rc.getRollingCountUnlocked()
+}
+
+func (rc *RollingCounter) OverThreshold() bool {
```
ElectroNafta: Should be guarded by an FF kill-switch.

imap/connectionlimiter/limiter.go
```diff
@@ -0,0 +1,141 @@
...
+	return &limiter{
+		limits:        limits,
+		sessionClient: make(map[int]Client),
+		clientCount:   make(map[Client]int),
+		log:           log,
+	}
+}
+
+func (l *limiter) TryBind(sessionID int, id imap.IMAPID) (allowed bool, key Client, current int, max int) {
+	l.mu.Lock()
```
ElectroNafta, line 52: Should be guarded by an FF kill-switch. We need a way to have all new connections succeed in that case.

imap/connectioncounter/connectioncounter.go
ElectroNafta: Rename `getRollingCountUnlocked` to `getRollingCountUnsafe`. This is the naming convention we follow for such cases. 

A method protected by a mutex on the other end would be `getRollingCountSafe`

imap/connectionlimiter/limiter.go
```diff
@@ -0,0 +1,141 @@
...
+	//TryBind tries to bind a given sessionID to a client.
+	TryBind(sessionID int, id imap.IMAPID) (allowed bool, key Client, current int, max int)
+
+	//Unbind unbinds a given sessionID from a client.
+	Unbind(sessionID int)
+}
+
+type limiter struct {
+	mu     sync.Mutex
+	limits Limits
```
ElectroNafta: As we discussed internally, we'd also like to be able to tweak these limits during runtime. 

Not the fullest extent, but we can have a fallbackLimits value that can be controlled by a feature flag. 

It will enable us to have a "pessimistic" and "optimistic" case, given the fact that we don't really have a definite number in mind. 

The limits that tryBind will use will be configured using that FF, the limits vs the fallback limits. 

Also add this as an option to the builder. And we'd need a nil check as well (if they're not configured)

imap/connectioncounter/connectioncounter_test.go
```diff
@@ -50,6 +50,7 @@ func (m *mockObsSender) LastValues() (int, int) {
```
ElectroNafta, line 49: Add another test case for the new field and rename the tests such that they correspond to the correct threshold we're testing.

imap/connectionlimiter/limiter.go
```diff
@@ -0,0 +1,151 @@
...
+		delete(l.clientCount, key)
+
+		l.log.WithFields(logrus.Fields{
+			"sessionID": sessionID,
+			"client":    key,
+		}).Debug("Unbinding session from client")
+	}
+}
+
+func (l *limiter) maxForKey(key Client, useFallback bool) int {
```
ElectroNafta, line 142: A comment would be worthwhile here explainign that if the key is not found in the limits we will return the default limit (non-blocking ~= 0)

imap/connectionlimiter/limits.go
```diff
@@ -0,0 +1,34 @@
+package connectionlimiter
+
+const (
+	defaultClientLimit  = 25
+	defaultUnknownLimit = 20
```
ElectroNafta: Should be set to 0 i.e. do not impose a limit on connections. If the fallback values

server.go
```diff
@@ -224,6 +228,10 @@ func (s *Server) Serve(ctx context.Context, l net.Listener) error {
 func (s *Server) serve(ctx context.Context, connCh <-chan net.Conn) {
 	connWG := async.MakeWaitGroup(s.panicHandler)
```
ElectroNafta, line 230: This would require a full bridge re-start in order for the changes to take place. The FF shouldn't be placed here. Just place it in useConnection limiter so we re-check it each time an event comes through.

ElectroNafta:
Good for the most part, just some minor things I've flagged. Let me provide you with proper FF values, and you can proceed with the changes.

imap/connectionlimiter/limiter.go
annatar: this is using the UnknownLimit which depends on the case, could be 0. could be something else. The unknown limit is the limit for `Unknown` Clients in this case

imap/connectionlimiter/limits.go
annatar: I will let these be the default, then create a function that returns `NewDefaultFallbackValues` which will be non blocking

option.go
```diff
@@ -278,10 +279,11 @@ func (w withConnectionRollingCounter) config(builder *serverBuilder) {
...
 }
 
-func WithConnectionRollingCounter(newConnectionTreshold, numberOfBuckets int, thresholdCheckInterval time.Duration) Option {
+func WithConnectionRollingCounter(observabilityThreshold, limitThreshold, numberOfBuckets int, thresholdCheckInterval time.Duration) Option {
 	return &withConnectionRollingCounter{
 		rollingConnectionCounter: connectioncounter.NewRollingCounter(
-			newConnectionTreshold,
+			observabilityThreshold,
+			limitThreshold,
 			numberOfBuckets,
```
ElectroNafta, line 287: Order is wrong:
`func NewRollingCounter(connectionLimitThreshold, observabilityConnectionThreshold, numberOfBuckets int, bucketRotationInterval time.Duration) *RollingCounter {
`

internal/session/session.go
```diff
@@ -355,3 +358,21 @@ func (s *Session) decodeMailboxName(name string) (string, error) {
 
 	return utf7.Encoding.NewDecoder().String(fmt.Sprintf("INBOX%v%v", delimiter, split[1]))
 }
+
+func (s *Session) CloseWithBye(reason string) error {
+	var retErr error
+	s.closeOnce.Do(func() {
+		s.writeMu.Lock()
+		defer s.writeMu.Unlock()
```
ElectroNafta: writeMu is only used here, and closeOnce should not let any other goroutines through, we can eliminate it.

internal/unleash/featureflags/flags.go
```diff
@@ -1,6 +1,9 @@
 package featureflags
 
 const (
-	CommandWatcherGlobalDisabled         = "InboxBridgeGenericImapOkHeartbeatDisabled"
-	CommandWatcherNonThunderbirdDisabled = "InboxBridgeGenericImapOkHeartbeatNonThunderbirdDisabled"
+	CommandWatcherGlobalDisabled             = "InboxBridgeGenericImapOkHeartbeatDisabled"
+	CommandWatcherNonThunderbirdDisabled     = "InboxBridgeGenericImapOkHeartbeatNonThunderbirdDisabled"
+	DisableConnectionLimiter                 = "InboxBridgeGluonConnectionLimiterDisabled"
+	ConnectionLimiterUseFallback             = "InboxBridgeGluonConnectionLimiterDefaultLimitsDisabled"
+	DisableConnectionCounterConnectionsLimit = "InboxBridgeGluonRollingCounterConnectionLimitDisabled"
```
ElectroNafta: Please follow the Kill-Switch naming convention here, disabled should be placed at the end. All of the newly introduced FFs  are kill-switches.

server.go
```diff
@@ -265,6 +280,46 @@ func (s *Server) serve(ctx context.Context, connCh <-chan net.Conn) {
 	}
 }
 
+func (s *Server) useConnectionLimiter(ctx context.Context, connWG *async.WaitGroup) {
+	disableConnectionLimiter := s.featureFlagProvider.GetFlagValue(featureflags.DisableConnectionLimiter)
```
ElectroNafta: This would still require a bridge restart for the FF changes to take place. 

We need to add it to lines 302-303. Whenever IMAPID comes in.

imap/connectionlimiter/limiter.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+
+	// already bound to this client, no-op allow
+	if prev, ok := l.sessionClient[sessionID]; ok && prev == key {
+		maxUsages := l.maxForKey(key, useFallback)
+		l.log.WithFields(logrus.Fields{
+			"sessionID": sessionID,
+			"client":    key,
+			"current":   l.clientCount[key],
+			"max":       max,
```
ElectroNafta: You're logging max, not maxUsages.

imap/connectionlimiter/limiter.go
```diff
@@ -0,0 +1,158 @@
...
+		if limit, ok := l.limits.PerClient[key]; ok {
+			return limit
+		}
+	}
+	l.log.WithFields(logrus.Fields{
+		"client": key,
+		"limit":  l.limits.UnknownLimit,
+	}).Debug("Client key not found in limits, using unknown limit")
+
+	return l.limits.UnknownLimit
```
ElectroNafta: Still I would reccomend to set the unkown limit to be 0 by default. You did make the change with the fallback ones, but here we're using the default.

annatar: Yes this should use the fallback unknown limit when using the fallback, but in my opinion it should not return a constant value, as the struct is configurable by default and I think this should respect the configuration passed by the consumer of this

ElectroNafta (approved):

PR #456, opened 2026-03-11
fix(BRIDGE-382): properly close context if Start gets called multiple times

annatar:
If gluon `Serve` is called twice on the same server instance RollingCounter overrides the context and context cancel, and since we use a RollingCounter in Bridge this causes a deadlock which can indefinitely block.

### DIFF
--- a/imap/connectioncounter/connectioncounter.go
+++ b/imap/connectioncounter/connectioncounter.go
@@ -91,6 +91,9 @@ func (rc *RollingCounter) run() {
 }
 
 func (rc *RollingCounter) Stop() {
+	if rc.cancel == nil {
+		return
+	}
 	rc.bucketRotationTicker.Stop()
 	rc.cancel()
 	rc.wg.Wait()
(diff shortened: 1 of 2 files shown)


imap/connectioncounter/connectioncounter.go
ElectroNafta, line 100:
```suggestion
func (rc *RollingCounter) Stop() {
if (rc.cancel == nil) {
   return
 }
rc.bucketRotationTicker.Stop()
rc.cancel()
rc.wg.Wait()
}
```

This way we can handle it in server.Serve() i.e.:
```go
func (s *Server) Serve(ctx context.Context, l net.Listener) error {
	ctx = observability.NewContextWithObservabilitySender(ctx, s.observabilitySender)
	ctx = reporter.NewContextWithReporter(ctx, s.reporter)
	ctx = contexts.NewDisableParallelismCtx(ctx, s.disableParallelism)

	s.publish(events.ListenerAdded{
		Addr: l.Addr(),
	})

	if s.connectionRollingCounter != nil {
		s.connectionRollingCounter.Stop()
		s.connectionRollingCounter.Start(ctx, s.observabilitySender, s)
	}
```

ElectroNafta:
It would be cleaner if we handle it over in server.Serve instead. That is the caller, and it makes the cancel explicit.

ElectroNafta (approved):

PR #459, opened 2026-03-17
feat(BRIDGE-492): update golangci-lint to v2 & fix linter issues

annatar:
Note: I've added the `unused` to be ignored in the tests

### DIFF
--- a/async/context.go
+++ b/async/context.go
@@ -28,6 +28,7 @@ func (a *Abortable) newCancelCtx(ctx context.Context) context.Context {
 	a.abortLock.Lock()
 	defer a.abortLock.Unlock()
 
+	//nolint: gosec //disable G118
 	ctx, cancel := context.WithCancel(ctx)
 
 	a.abortFunc = append(a.abortFunc, cancel)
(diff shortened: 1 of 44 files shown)


internal/session/handle_starttls.go
```diff
@@ -19,6 +19,7 @@ func (s *Session) handleStartTLS(tag string, _ *command.StartTLS) error {
 
 	conn := tls.Server(s.conn, s.tlsConfig)
 
+	//nolint:noctx
```
ElectroNafta: This isn't in the linters, why are we disabling it? Also handshake doesn't even accept ctx argument.

internal/db_impl/sqlite3/read_ops.go
```diff
@@ -508,6 +508,7 @@ func (r readOps) GetMessagesFlags(ctx context.Context, ids []imap.InternalMessag
 
 		args := utils.MapSliceToAny(chunk)
 
+		//nolint: unused
```
ElectroNafta, line 511: I don't understand how this even compiles if it's unused.

async/context.go
```diff
@@ -28,6 +28,7 @@ func (a *Abortable) newCancelCtx(ctx context.Context) context.Context {
 	a.abortLock.Lock()
 	defer a.abortLock.Unlock()
 
+	//nolint:gosec
```
ElectroNafta: Same as in the Bridge MR, please disable the particular rule instead of the linter.

ElectroNafta (approved):

PR #466, opened 2026-04-09
fix(BRIDGE-515): replace old go syntax using go:fix to match new go 1.26 syntax

annatar:


store/semaphore_test.go
```diff
@@ -44,9 +44,7 @@ func TestSemaphore(t *testing.T) {
 func wait(fn func()) func() {
 	var wg sync.WaitGroup
 
-	wg.Add(1)
-
-	go func() { defer wg.Done(); fn() }()
+	wg.Go(func() { ; fn() })
```
ElectroNafta: NIT: No need for the semicolon here

internal/backend/connector_updates.go
```diff
@@ -357,10 +357,7 @@ func (user *user) applyMessagesCreated(ctx context.Context, update *imap.Message
 
 		for _, chunk := range xslices.Chunk(messagesToCreate, db.ChunkLimit) {
 			// Create messages in the store in parallel
-			numStoreRoutines := runtime.NumCPU() / 4
-			if numStoreRoutines < len(chunk) {
```
ElectroNafta, line 361: The transformation is correct, but this doesn't make much sense. I don't think the intention here was to spawn as many go-routines as there are chunks but rather up to a limit of numStoreRoutines. 

We shouldn't handle this here, rather potentially open up a separate ticket.

tests/connection_test.go
```diff
@@ -30,21 +30,21 @@ func lines(lines ...string) string {
...
 	return res
 }
 
 func seq(begin, end int) string {
-	var res string
+	var res strings.Builder
 
 	for i := begin; i < end; i++ {
-		res += strconv.Itoa(i) + " "
+		res.WriteString(strconv.Itoa(i) + " ")
```
ElectroNafta: This allocates a new string to concatenate. Doing two res.WriteStrings() would be better.

tests/connection_test.go
```diff
@@ -30,21 +30,21 @@ func lines(lines ...string) string {
...
 	}
 
-	return res + strconv.Itoa(end)
+	return res.String() + strconv.Itoa(end)
```
ElectroNafta: Same applies here, just a single WriteStrings, and res.String() return. 

Both of these are nits since they affect a test helper utility.

internal/backend/connector_updates.go
annatar: I will open a separate ticket for this

tests/connection_test.go
annatar: Updated

tests/connection_test.go
annatar: Updated

ElectroNafta (approved):

